Issue #23, opened 2020-05-21
Proposal - project structure


mbraekman:
**Below is a suggestion for the structure of this repo.**
_Feedback is welcome_: @stijnmoreels, @tomkerkhove _(Tom: ignore this for the coming weeks!!)_ or anyone else reading this.

Since the purpose of this repo will not be limited to PowerShell, I suggest to start with the following structure:
   - /src/Arcus.Scripting.PowerShell/...
   - /docs/Arcus.Scripting.PowerShell/...

_Later on we can always add scripts for Bash,... depending on needs or suggestions_
 
As we are going to have several scripts related to multiple resource types/areas, I would like to suggest to create a PS-module per area:
  - Arcus.Scripting.PS.ApiManagement
  - Arcus.Scripting.PS.ARM
  - ...

This would prevent someone from having to pull all scripts while only those related to DataFactory, for instance, are required. We can also easily push new scripts or fixes related to a single area without impacting any others.

This means that for each of these area's, we will need a separate project (incl. psm1 & psd1), separate build/release and thus increasing the overhead.

Is this overhead acceptable for you? 
Do you agree with the proposed structure?
Any other suggestions?

Thanks for the feedback!

stijnmoreels:
Yes, I can agree on the splitting by area/domain. That seems appropriate.

I would suggest for the docs, to not add this in the paths and more create a structue like the other Arcus projects:

* `./docs/features/arm.md
* `./docs/features/api-management.md`

And for preview and older versions:
* `./docs/preview/features/`
* `./docs/v0.1.0/features/`

Is this something you would agree to? Then we would have the same docs strucure here.

mbraekman:
I like the idea of using a uniform structure across all projects, so your proposal for using the preview and version-folders, as is the case for the other projects, is fine by me!

I would like to make a separation based on the language as well. Even though it might seem overkill now and we might never need it, if we would ever expand do different languages we don't need any refactoring. 
Because I can see that someone who is interested in the PowerShell-documentation, wouldn't want to see any bash-related docs within the same page.

stijnmoreels:
I'm currently working on a this proposal and the CI build so we have some 'starter' project and can begin with the actual implementation.

tomkerkhove:
That's ok for me, but I would rename `ps` with `powershell` for sake of clarity.

mbraekman:
Wanted to keep it short, but ok, might not be as clear -> agree with full name

[...]

PR #32, opened 2020-05-29
Feature - add initial project structure

### LINKED ISSUE #25: Configure CI build
onfigure a CI build that will build & test the changes made in every PR

### LINKED ISSUE #29: Provide a release pipeline for continuous deployment system (maybe MyGet?)
Provide a release pipeline for continuous deployment system
      
### Checklist
      
- [x] Configure to run for every PR
- [x] Use 'Determine NuGet Package Version' task group to determin package version
- [x] Build the codebase
- [x] Run test suite
- [ ] Push all modules to continuous deployment system (maybe MyGet?)

stijnmoreels:
Relates #23 
Closes #29 
Closes #25


build/ci-build.yml
```diff
@@ -111,3 +109,20 @@ stages:
               artifact: 'Build'
               path: '$(Build.SourcesDirectory)'
           - template: nuget/publish-preview-package.yml@templates
+            parameters:
```
stijnmoreels, line 112: We'll have to see how we could maybe publish PowerShell modules to the MyGet source, but I'll have to wait for @tomkerkhove bc I think the service connection is still on his account and not in the shared Lastpass group.

mbraekman: Not sure whether this is to be published onto MyGet though, as this would mean publishing the modules on both MyGet and into the PowerShell-gallery.
I would stick to the PowerShell-gallery, which also offers the possibility doing a [pre-release](https://docs.microsoft.com/en-us/powershell/scripting/gallery/concepts/module-prerelease-support?view=powershell-7) for validation purposes.

But perhaps I'm missing some of the capabilities of MyGet, as my experience with it is limited, so feel free to correct me.

stijnmoreels: Yeah, we had a MyGet release for the CI build so we could already test the deployment continuously. This is done for every repo in Arcus. I don't know actually if it's an added value here.

As far as support, I thought PS publishing was supported: https://docs.myget.org/docs/how-to/publish-a-powershell-module-to-myget
But we can maybe hold of on this? No problem.

mbraekman: Since it's done across all Arcus repo's, perhaps we can do it as well to keep it alligned, but lets hold of on this for a while untill we can verify with Tom whether this would be beneficial for this project.

Thinking about it again, it would offer us a "private" store to verify the modules before publising them onto the PowerShellGallery, increasing quality of the publicly available modules.
Just wondering whether this will cause version-numbers being skipped on the PS-gallery then?
For isntance, if v1.0.1 is published on MyGet first -> a bug is discovered + fixed -> v1.0.2 is published, then we'll go from v1.0.0 to v1.0.2 on the PS-gallery. Would be nice to avoid these gaps, but perhaps this is nitpicking..

stijnmoreels: > Since it's done across all Arcus repo's, perhaps we can do it as well to keep it alligned, but lets hold of on this for a while untill we can verify with Tom whether this would be beneficial for this project.

Yes, indeed we can maybe leave it out here and come back later.

stijnmoreels: > Thinking about it again, it would offer us a "private" store to verify the modules before publising them onto the PowerShellGallery, increasing quality of the publicly available modules.
Just wondering whether this will cause version-numbers being skipped on the PS-gallery then?
For isntance, if v1.0.1 is published on MyGet first -> a bug is discovered + fixed -> v1.0.2 is published, then we'll go from v1.0.0 to v1.0.2 on the PS-gallery. Would be nice to avoid these gaps, but perhaps this is nitpicking..

The versioning will be on a different structure then the 'actual' releases. We're using the Arcus build functionality here that allows us to determine either a 'PR version' or a 'release version' for our package. So to answer your question, no, they're will not be versions left out bc it actually is two total different pipelines with different purposes. It's like you said, for having a constantly private preview of the library. 
But again, let's wait for Tom.
Thanks for thinking about this!

stijnmoreels: The MyGet stage is now present but is skipped, so we have it still in source control for later use but are ignoring it now. Is good for you, or you want it removed here?

mbraekman: The MyGet stage can be kept in the file, untill we make a final decision about this.

Issue #9, opened 2020-05-12
Provide script to get all access policies for an Azure Key Vault


tomkerkhove:
Provide script to get all access policies for an Azure Key Vault

mbraekman:
Based on the following script:
https://codit.visualstudio.com/Invictus%20Templates/_git/invictus-templates?path=%2Fdeploy%2FInfra-GetKeyVaultAccessPolicies.ps1

PR #43, opened 2020-06-08
Feature - add 'Get KeyVault access policies' script

stijnmoreels:
- Added new module project with script
- Added Pester unit test with mocked out Az function

Relates to #9


docs/preview/features/ps/keyvault.md
```diff
@@ -0,0 +1,38 @@
+---
+title: "Scripts related to interacting with Azure Key Vault"
+layout: default
+---
+
+# Azure Key Vault scripting
```
tomkerkhove, line 6: I think it's ok to leave this out
```suggestion
# Azure Key Vault
```

docs/preview/features/ps/keyvault.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Installation
+
+To have access to the following features, you have to import the module:
+
+```powershell
+PS> Import-Module -Name Arcus.Scripting.PS.KeyVault
```
tomkerkhove, line 13: Is there a reason why we have `PS` in here? Do we support others?

docs/preview/features/ps/keyvault.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Get all access policies of a key vault
+
+Lists the current available access policies of the Azure key vault resource into a new variable availble at runtime during a Azure DevOps build.
+
+| Parameter            | Mandatory                                     | Description                                                                  |
+| -------------------- | --------------------------------------------- | ---------------------------------------------------------------------------- |
+| `KeyVaultName`       | yes                                           | The name of the key vault from which the access policies are to be retrieved |
+| `ResourceGroupName`  | no                                            | The resource group containing the key vault                                  |
+| `OutputVariableName` | no (default: `Infra.KeyVault.AccessPolicies`) | The name of the variable to be added to DevOps-pipeline variables at runtime |
```
tomkerkhove, line 24: Maybe it's best to drop the default and make it mandatory? We should not leak our internal convention to others maybe?

Just a thought.

docs/preview/index.md
```diff
@@ -0,0 +1,28 @@
+---
+title: "Home"
+layout: default
+permalink: /
+redirect_from:
+ - /index.html
+---
+
+[![NuGet Badge](https://buildstats.info/nuget/Arcus.{Component}?includePreReleases=true)](https://www.nuget.org/packages/Arcus.{Component}/)
```
tomkerkhove, line 9: NuGet is not relevant here.

docs/preview/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Installation
+
+The Arcus.{Component} can be installed via NuGet:
+
+```shell
+PM > Install-Package Arcus.{Component}
```
tomkerkhove, line 16: We should update this

tomkerkhove: See master

docs/preview/features/ps/keyvault.md
```diff
@@ -0,0 +1,38 @@
...
+
+**Example**
+
+```powershell
+PS> Get-KeyVaultAccessPolicies -KeyVaultName "my-key-vault"
+# Output: #vso[task.setvariable variable=Infra.KeyVault.AccessPolicies]{list: [{tenantId: ...,permissions: ...}]}
```
tomkerkhove, line 30: This is confusing - Does the command set this as an output variable or also write it? Shouldn't it just return the payload and you can capture it similar to this?

```powershell
$acls = Get-KeyVaultAccessPolicies -KeyVaultName "my-key-vault" -ResourceGroupName "my-resouce-group" -OutputVariableName "My.KeyVault.AccessPolicies"
```

And then do this:
```powershell
#vso[task.setvariable variable=My.KeyVault.AccessPolicies] $acls
```

Because it feels like we are mixing two things here.

docs/preview/features/ps/keyvault.md
mbraekman: Currently: no
This was included to ensure the repo would be future-proof and while those might not end up on the PS-gallery, seems better to allign naming conventions?
Related to #23

docs/preview/features/ps/keyvault.md
```diff
@@ -0,0 +1,38 @@
...
+
+To have access to the following features, you have to import the module:
+
+```powershell
+PS> Import-Module -Name Arcus.Scripting.PS.KeyVault
+```
+
+## Get all access policies of a key vault
+
+Lists the current available access policies of the Azure key vault resource into a new variable availble at runtime during a Azure DevOps build.
```
mbraekman, line 18: small typo: 
- availble -> available
- a Azure DevOps -> an Azure DevOps

docs/preview/features/ps/keyvault.md
mbraekman: I see, perhaps best to leave this out as suggested.

docs/preview/features/ps/keyvault.md
tomkerkhove: Yes, I think it's best to leave it out indeed as different scripting might require different package managers indeed

docs/preview/features/ps/keyvault.md
tomkerkhove, line 18:
```suggestion
Lists the current available access policies of the Azure key vault resource into a new variable available at runtime during automated processes such as an Azure DevOps build.
```

tomkerkhove: Agreed, extended it in a proposal below to not only pin it to Azure DevOps.

docs/preview/features/ps/keyvault.md
mbraekman: This script was ment to capture the current access policies and set it as an pipeline variable to be passed along to an ARM-template.

Either we rename this command  and create to second:
- Get-KeyVaultAccessPolicies: actually outputs the value
- Write-KeyVaultAccessPoliciesToDevOpsVariable: adds the variable
(or some better naming :) )

We could expect the user to use `#vso[task.setvariable variable=My.KeyVault.AccessPolicies] $acls` themselves, but this would add some additional "complexity", while this should be all about making it easier.

tomkerkhove: What if that `Write-KeyVaultAccessPoliciesToDevOpsVariable` simply becomes `Set-DevOpsVariable` instead? That way you can re-use it for others as well.

mbraekman: We can definitely use a script such as `Set-DevOpsVariable` (#44 has been created for this)

Makes sense, we'll just have to provide clear documentation that this command no longer writes the output into a pipeline variable - as this is now the case, people might still expect it to do this or complain their setup is no longer working after migrating to this module.

docs/preview/features/ps/keyvault.md
stijnmoreels: See other conv: removed variable output. Resolving.

docs/preview/features/ps/keyvault.md
stijnmoreels: Moved `/ps` to `/powershell`.

docs/preview/index.md
stijnmoreels: Switched to PS gallery.

docs/preview/index.md
stijnmoreels: Merged with master and inserted correct info.

docs/preview/features/ps/keyvault.md
stijnmoreels: Added proposal below.

docs/preview/features/ps/keyvault.md
stijnmoreels: Removed the writing to the vso output variable and updated the test.

stijnmoreels: Although, I'm not sure what's the purpose of this line:

```powershell
$armAccessPoliciesParameter = [pscustomobject]@{
    list = $armAccessPolicies
}
```

stijnmoreels: If it doesn't add any value, I'll remove this and return the list (`$accessPolicies`) itself as return value.

tomkerkhove: I'll leave it up to @mbraekman

src/Arcus.Scripting.PS.KeyVault/Scripts/Get-Infra-KeyVaultAccessPolicies.ps1
```diff
@@ -0,0 +1,57 @@
+param(
```
tomkerkhove, line 1: I would rename this script to `Get-KeyVaultAccessPolicies` to be honest, `Infra` is legacy from our internal system if you ask me.

docs/preview/features/powershell/keyvault.md
```diff
@@ -0,0 +1,35 @@
...
+---
+
+# Azure Key Vault
+
+## Installation
+
+To have access to the following features, you have to import the module:
+
+```powershell
+PS> Import-Module -Name Arcus.Scripting.PS.KeyVault
```
tomkerkhove, line 13: Let's remove all of these `.PS` prefixes as discussed
```suggestion
PS> Import-Module -Name Arcus.Scripting.KeyVault
```

src/Arcus.Scripting.PS.KeyVault/Scripts/Get-Infra-KeyVaultAccessPolicies.ps1
tomkerkhove: Or should we align and use `Get-AzKeyVaultAccessPolicies` similar to the official cmdlets? Maybe not a bad thing?

src/Arcus.Scripting.PS.KeyVault/Arcus.Scripting.PS.KeyVault.psm1
```diff
@@ -0,0 +1,22 @@
+<#
+ .Synopsis
+  Return the current access policies present in KeyVault.
+
+ .Description
+  List the current access policies present in KeyVault.
+
+ .Parameter KeyVaultName
+  The name of the KeyVault from which the access policies are to be retrieved.
```
tomkerkhove, line 9: It's a product, so let's use the official name
```suggestion
  The name of the Azure Key Vault from which the access policies are to be retrieved.
```

src/Arcus.Scripting.PS.KeyVault/Arcus.Scripting.PS.KeyVault.psm1
```diff
@@ -0,0 +1,22 @@
...
+
+ .Parameter ResourceGroupName
+  The resource group containing the KeyVault.
```
tomkerkhove, line 12: It's a product, so let's use the official name
```suggestion
  The resource group containing the Azure Key Vault.
```

src/Arcus.Scripting.PS.KeyVault/Arcus.Scripting.PS.KeyVault.psm1
```diff
@@ -0,0 +1,22 @@
+<#
+ .Synopsis
+  Return the current access policies present in KeyVault.
+
+ .Description
+  List the current access policies present in KeyVault.
```
tomkerkhove, line 6: It's a product, so let's use the official name

```suggestion
  List the current access policies present in Azure Key Vault.
```

src/Arcus.Scripting.PS.KeyVault/Arcus.Scripting.PS.KeyVault.psm1
```diff
@@ -0,0 +1,22 @@
+<#
+ .Synopsis
+  Return the current access policies present in KeyVault.
```
tomkerkhove, line 3: It's a product, so let's use the official name

```suggestion
  Return the current access policies present in Azure Key Vault.
```

docs/preview/features/powershell/keyvault.md
stijnmoreels: Removed all `PS` path's in project.

src/Arcus.Scripting.PS.KeyVault/Scripts/Get-Infra-KeyVaultAccessPolicies.ps1
stijnmoreels: Renamed this.

docs/preview/index.md
```diff
@@ -0,0 +1,28 @@
+---
+title: "Home"
+layout: default
+permalink: /
+redirect_from:
```
tomkerkhove, line 5: Can you remove this please?

docs/preview/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+```powershell
+PS> Install-Module -Name Arcus.Scripting.ARM
+```
+
+For more granular packages we recommend reading the documentation.
+
+# Features
+
+* Azure key vault scripts ([powershell](features/powershell/key-vault))
```
tomkerkhove, line 23:
```suggestion
* Automate Azure Key Vault tasks ([powershell](features/powershell/key-vault))
```

docs/preview/index.md
```diff
@@ -0,0 +1,28 @@
...
+
+# License
+This is licensed under The MIT License (MIT). Which means that you can use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the web application. But you always need to state that Codit is the original author of this web application.
+
+*[Full license here](https://github.com/arcus-azure/arcus.{Component}/blob/master/LICENSE)*
```
tomkerkhove, line 28:
```suggestion
*[Full license here](https://github.com/arcus-azure/arcus.scripting/blob/master/LICENSE)*
```

src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
```diff
@@ -0,0 +1,57 @@
...
+	$armAccessPoliciesParameter = [pscustomobject]@{
+		list = $armAccessPolicies
+	}
+
+	Write-Host "Current access policies: $armAccessPoliciesParameter"
+	return $armAccessPoliciesParameter
+}
+else
+{
+	Write-Host "KeyVault '$keyVaultName' could not be found."
```
tomkerkhove, line 56: Write-Host or Write-Error?

src/Arcus.Scripting.KeyVault/Arcus.Scripting.KeyVault.psm1
```diff
@@ -0,0 +1,22 @@
...
+#>
+function Get-KeyVaultAccessPolicies {
+	param(
+	   [parameter(Mandatory = $true)][string] $KeyVaultName,
+	   [parameter(Mandatory = $false)][string] $ResourceGroupName = ""
+	)
+	. $PSScriptRoot\Scripts\Get-AzKeyVaultAccessPolicies.ps1 -keyVaultName $KeyVaultName -resourceGroupName $ResourceGroupName
+}
+
+Export-ModuleMember -Function Get-KeyVaultAccessPolicies
```
tomkerkhove, line 22: Is this correct or do we need the Az prefix here as well?

src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
mbraekman: Could go either way:
- On the initial execution of a pipeline, the KV doesn't exist yet, so you cannot retrieve the access policies, but you don't want the command to fail in this case.
- On the other side, trying to retrieve access policies of a non-existing KV is correct to fail..

Perhaps a better option would be:
- add a switch-parameter to indicate whether it should `Write-Host` or `Write-Error` (default = `Write-Error`)

src/Arcus.Scripting.KeyVault/Arcus.Scripting.KeyVault.psm1
mbraekman: Internal and external command-names should be alligned.

Biggest question is, would you refer to the Az-module in the name of the commands? I don't think we'll support an overload using the AzureRm-module?

docs/preview/index.md
stijnmoreels: Yes, done.

src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
stijnmoreels: Hmm, shouldn't that be the job of the common parameter, `-ErrorAction`? Couldn't we use always `WriteError` and let the consumer decide with `ErrorAction` and `ErrorVariable` how errors should be processed?

src/Arcus.Scripting.KeyVault/Arcus.Scripting.KeyVault.psm1
stijnmoreels: So the file name has been added with the`Az` prefix but both the internal and external commds have not. Is that okay here to resolve this?

tomkerkhove: I would use Az as a reference to Azure in general so that it's clear we are talking about Azure API Management and not API Management in general for example.

I'd take it all the way and include it in file names, but that's just my opinion!

src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
tomkerkhove: You are right @mbraekman, didn't think of the first scenario; in that case I would use `Write-Warning` to be explicit but that it can be ok - Thoughts?

mbraekman: Let's go for `Write-Warning` by default and let the consumer decide otherwise using `-ErrorAction`, as suggested by Stijn.

src/Arcus.Scripting.KeyVault/Arcus.Scripting.KeyVault.psm1
mbraekman: Good point, didn't think outside of the "Azure-box" here.

Let's go for consistency and add the `Az`-prefix all over the place: internal command, external command and filename.

src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
stijnmoreels: Ok, changed that.

src/Arcus.Scripting.KeyVault/Arcus.Scripting.KeyVault.psm1
stijnmoreels: Ok, did that.

docs/preview/features/ps/keyvault.md
stijnmoreels: > I'll leave it up to @mbraekman

Any thoughts @mbraekman ?

docs/preview/index.md
```diff
@@ -0,0 +1,25 @@
+---
+title: "Home"
+layout: default
+---
+
+![PowerShell Gallery Version](https://img.shields.io/powershellgallery/v/Arcus.Scripting.ARM)
+
+# Installation
+
+The Arcus.Scripting can be installed via NuGet:
```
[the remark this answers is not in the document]
stijnmoreels: Woops indeed, thanks.

stijnmoreels:
> Just some small remarks.
> Regarding the `Az`-prefix, remember to include the Az-prefix in the Arcus.Scripting.KeyVault.psd1-file.

Ah yes, thanks, just added this.

docs/preview/features/ps/keyvault.md
mbraekman: The reason for that additional object is to ensure that the output-structure is valid to be used as a parameter-value of an ARM-template.
ARM expects the following structure, which is achieved with the additional conversion:
```
{
  "list": [
     { "tenantId": "123456", ... }
  ]
}
```

While the default structure is:
```
[
  { "tenantId": "123456", ... }
]
```

stijnmoreels: Ok, then I'll leave it like this.

Issue #30, opened 2020-05-29
Provide a release pipeline for Powershell gallery


stijnmoreels:
Provide a release pipeline for NuGet.org.
      
### Checklist
      
- [x] Build the codebase
- [x] Run test suite
- [x] Tag the codebase on success
- [x] Create a GitHub pre-release for a preview releases
- [x] Create a GitHub release for full releases
- [x] Push all Powershell cmdlet to Powershell gallery

stijnmoreels:
Leave open for service connection that has to be made to PowerShell Gallery.

tomkerkhove:
Already done ;)

PR #61, opened 2020-07-02
Feature - publish to PS Gallery

stijnmoreels:
Relates to #30


src/Arcus.Scripting.ApiManagement/Arcus.Scripting.ApiManagement.psd1
```diff
@@ -48,7 +48,7 @@ Description = 'Scripts related to Azure API Management'
 # ProcessorArchitecture = ''
 
 # Modules that must be imported into the global environment prior to importing this module
-RequiredModules = @('Az.ApiManagement')
+RequiredModules = @("Az.ApiManagement")
```
stijnmoreels, line 51: Apperently, adding these required modules to the manifest throws an exception I can't yet figure out how to solve. Any ideas?

tomkerkhove: Maybe it's the format? Azure PowerShell does it like this, but not sure if it's custom or not - https://github.com/Azure/azure-powershell/blob/master/src/ApiManagement/ApiManagement/Az.ApiManagement.psd1#L56

build/psgallery-release.yml
```diff
@@ -88,12 +89,18 @@ stages:
...
+          #    repositoryName: '$(Repository)'
+          #    releaseNotes: |
+          #      Install new version via [PowerShell Gallery](https://www.powershellgallery.com/packages?q=$(Project)
+          #      ```shell
+          #      PS > Install-Module -Name $(Project).* --Version $(Build.BuildNumber)
+          #      ```
+          - powershell: |
+              Import-Module PowerShellGet -Force
+              Get-ChildItem -Path ./src -Filter *.psm1 -Recurse |
+                  % { Publish-Module -Path $_.DirectoryName -NuGetApiKey $env:APIKEY -Repository PSGallery -Verbose -Force }
```
tomkerkhove, line 103: Instead of `$env:APIKEY`, why are we not using typical variables?

tomkerkhove: I see you pass them as env var but think you can just consume them directly 🤔

stijnmoreels: Yeah, I tried that in the past and I always come back to this bc I couldn't have access to them directly. I think even the official docs explains this in some parts that you have to pass it allong as env vars. Maybe bc it's inline.

src/Arcus.Scripting.ApiManagement/Arcus.Scripting.ApiManagement.psd1
stijnmoreels: Yeah, I tried different formats. Normally just a string just be correct too. I changed it as in the link you send... same result 😭 . Not sure what I'm doing wrong here.

tomkerkhove: What I would do to improve our modules is:
- Include `PowerShellVersion` (see [sample](https://github.com/Azure/azure-powershell/blob/master/src/ApiManagement/ApiManagement/Az.ApiManagement.psd1#L38))
- Add `Tags`, `LicenseUri`, `ProjectUri`, `IconUri` in private data (see [sample](https://github.com/Azure/azure-powershell/blob/master/src/ApiManagement/ApiManagement/Az.ApiManagement.psd1#L186))

tomkerkhove: What if we try this?
```suggestion
RequiredModules = @(@{ModuleName = 'Az.ApiManagement'; ModuleVersion = '2.0.1'; })
```

tomkerkhove: However, I would add the info mentioned above as well when testing, maybe it's because of the combination (min version + module)

tomkerkhove:
So [Az is a reserved name

tomkerkhove:
If that doesn't work; we will have a look at how Azure PowerShell does things

stijnmoreels:
> If that doesn't work; we will have a look at how Azure PowerShell does things

Maybe we can use NuGet directly, instead of relying on PowerShellGet and create the package ourselves.
* [Create nuspec from psd1 file](https://gist.github.com/tyconsulting/c567e5cc66fc522d46743d744579be27)
* Publish the package to `https://www.powershellgallery.com/api/v2/package/`

stijnmoreels:
> > If that doesn't work; we will have a look at how Azure PowerShell does things
> 
> Maybe we can use NuGet directly, instead of relying on PowerShellGet and create the package ourselves.
> 
> * [Create nuspec from psd1 file](https://gist.github.com/tyconsulting/c567e5cc66fc522d46743d744579be27) (This is extracted from the PowerShellGet repo itself)
> * Publish the package to `https://www.powershellgallery.com/api/v2/package/`

Who! I think I found it.
image

build/tools/psd1-to-nuspec.ps1
```diff
@@ -0,0 +1,549 @@
...
+  #variables
+  $Name = $null
+  $Description = $null
+  $Version = ''
+  $Author = $null
+  $CompanyName = $null
+  $Copyright = $null
+  $requireLicenseAcceptance = 'false'
+ 
+  $PSModuleInfo = Test-ModuleManifest -Path $ManifestPath -ErrorAction SilentlyContinue
```
stijnmoreels, line 240: Now, I did skip the failures in the `Test-ModuleManifest` though.

stijnmoreels: @tomkerkhove , do you want me to write some extra logic to comment out the `RequiredModules` in the `.psd1` files before we run this and then uncomment them afterwards?
It defeats the purpose a bit by doing so, but then we can fail on other validation failures.

src/Arcus.Scripting.ApiManagement/Arcus.Scripting.ApiManagement.psd1
tomkerkhove: https://github.com/arcus-azure/arcus.scripting/issues/65

tomkerkhove:
Looks like we are deleting `src/Arcus.Scripting.ApiManagement/Arcus.Scripting.ApiManagement.psd1`, is that on purpose? How do we define metadata then?

build/tools/psd1-to-nuspec.ps1
tomkerkhove: If we can and you are up for it, I would give that a try for sure! Better have validation rather than none

build/psgallery-release.yml
```diff
@@ -16,6 +16,7 @@ resources:
 
 variables:
   - group: 'Build Configuration'
+  - group: 'PowerShell Gallery'
```
tomkerkhove, line 19: Do we still need this since we use a Service Connection? If not, let's remove the variable group as well

build/psgallery-release.yml
```diff
@@ -92,8 +93,26 @@ stages:
             parameters:
               repositoryName: '$(Repository)'
               releaseNotes: |
-                Install new version via [NuGet](https://www.nuget.org/packages/$(Project)/$(Build.BuildNumber))
+                Install new version via [PowerShell Gallery](https://www.powershellgallery.com/packages?q=$(Project)
                 ```shell
-                PM > Install-Package $(Project) --Version $(Build.BuildNumber)
+                PS > Install-Module -Name $(Project).* --Version $(Build.BuildNumber)
```
tomkerkhove, line 98: Let's not use a `.*` but a concrete example that they can run maybe? For example `Arcus.Scripting.ARM` to align with our README.

tomkerkhove: This makes me wonder, maybe another opportunity for a `Arcus.Scripting` metapackage @mbraekman ?

build/tools/psd1-to-nuspec.ps1
```diff
@@ -0,0 +1,564 @@
+#Requires -Modules PowerShellGet
+#Requires -Version 5.0
+<#
+    =======================================================================================================
+    AUTHOR:  Tao Yang
```
tomkerkhove, line 5: 😏

build/tools/psd1-to-nuspec.ps1
```diff
@@ -0,0 +1,564 @@
+#Requires -Modules PowerShellGet
+#Requires -Version 5.0
+<#
+    =======================================================================================================
+    AUTHOR:  Tao Yang 
+    DATE:    05/09/2018
```
tomkerkhove, line 6: 😏

build/tools/psd1-to-nuspec.ps1
tomkerkhove: Maybe add a reference to the location where this was copied from in the comment?

build/tools/psd1-to-nuspec.ps1
```diff
@@ -0,0 +1,564 @@
...
+  {
+    if($NuspecPath -and (Test-Path -Path $NuspecPath -PathType Leaf) -and (Get-Content -LiteralPath $NuspecPath -Raw))
+    {
+      Write-Output -InputObject $NuspecPath
+    }
+  }
+}
+#endregion
+
+#region main
```
tomkerkhove, line 547: Let's remove the regions 😄

build/tools/psd1-to-nuspec.ps1
tomkerkhove: Ha, I think this is already done here: https://github.com/arcus-azure/arcus.scripting/pull/61/files#diff-75b200261f61d0658f8ca326c14b6373R252

build/psgallery-release.yml
mbraekman: Could be usefull, although most of the times people will only be interested in a specific module, I suppose.
Let's add a task for this to be implemented in one of the future versions.

tomkerkhove: That's a good point!

mbraekman: #66 (and similar #67) has been created for this.

build/tools/psd1-to-nuspec.ps1
stijnmoreels: 😄 , yes, I implemented it just yesterday.

build/tools/psd1-to-nuspec.ps1
stijnmoreels: Ok.

stijnmoreels:
> Can we add release notes in GH and add the URL to the release notes?

What do you mean? Adding release notes in the GitHub release? (done via the YAML DevOps template) Adding release notes to the published PS modules? (done via the metadata in each PS module project).

mbraekman:
> > Can we add release notes in GH and add the URL to the release notes?
> 
> What do you mean? Adding release notes in the GitHub release? (done via the YAML DevOps template) Adding release notes to the published PS modules? (done via the metadata in each PS module project).

In the psd1-file you can provide released notes or a link to the release notes. Would be nice if this would directly link to the generated GitHub release notes.

stijnmoreels:
> > > Can we add release notes in GH and add the URL to the release notes?
> > 
> > 
> > What do you mean? Adding release notes in the GitHub release? (done via the YAML DevOps template) Adding release notes to the published PS modules? (done via the metadata in each PS module project).
> 
> In the psd1-file you can provide released notes or a link to the release notes. Would be nice if this would directly link to the generated GitHub release notes.

Ok, we can do it in another PR (since this one is actually about the publish release itself) #70

Issue #67, opened 2020-07-13
Create metapackage: Arcus.Scripting.Storage


mbraekman:
**Is your feature request related to a problem? Please describe.**
Create metapackage: Arcus.Scripting.Storage

**Describe the solution you'd like**
Create 1 package which imports all om the Arcus.Scripting.Storage.* modules.

PR #77, opened 2020-07-29
Feature - move Table Storage to nested module

### LINKED ISSUE #72: Rename Arcus.Scripting.Storage to Arcus.Scripting.Storage.Table
Create a specific module for table-storage operations.
As was discussed in #54 

The current module Arcus.Scripting.Storage only contains operations related to Table Storage.
Rename this module to reflect this, and in the future a new overlapping module Arcus.Scripting.Storage will be created, as mentioned in #67

stijnmoreels:
Closes #72


stijnmoreels:
Hmm, no that suggestion of 'just' renaming doesn't seem to work if you want to import it without the nested name.

mbraekman:
What would be required is to rename this entire module to become `Arcus.Scripting.Storage.Table`.
(Nothing should be left behind with the name `Arcus.Scripting.Storage`)

Later on, we'll create a separate module, which will act as a "parent"-module (see #67), in which we will only add a psd1-file setting the `NestedModules`-property to import all storage-related modules, being `Arcus.Scripting.Storage.Table` and for instance `Arcus.Scripting.Storage.Blob` (once it would be created).

PR #127, opened 2021-01-27
Chore - validate PowerShell module manifest during CI build

stijnmoreels:
Validation of all the Arcus PowerShell modules during the CI build pipeline.

Closes #126

PR #130, opened 2021-01-28
Fix - push PS modules with `RequiredModules` to MyGet

### LINKED ISSUE #128: PS modules with `RequiredModules` don't get pushed to MyGet
**Describe the problem**
Because the `Publish-Module` uses `Test-ModuleManifest`, it not only validates syntax but required modules too. This results in a failed push attempt.

**How to resolve?**
We could install all modules that's required before practice pushing, but that's a bit unneeded overhead. We could investigate in pushing the NuGet package instead like we do in the release pipeline, so it's only validated by syntax.

stijnmoreels:
Publish PS modules with the low-level `nuget` CLI commands so we can skip the `RequiredModules` validation which is not correct in this context. An extra module manifest syntax validation is happening (#127).

Closes #128

### DIFF
--- a/build/ci-build.yml
+++ b/build/ci-build.yml
@@ -121,14 +121,14 @@ stages:
             inputs:
               artifact: 'Build'
               path: '$(Build.SourcesDirectory)'
-           - powershell: |
+          - powershell: |
               Import-Module PowerShellGet -Force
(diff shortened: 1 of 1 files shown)


mbraekman (approved):
Great!

PR #125, opened 2021-01-27
Feature - create Azure Storage metapackage

stijnmoreels:
Creates new PS module project that imports all the Arcus Azure Storage PS module projects.

Closes #67

PR #132, opened 2021-01-28
Feature - create All metapackage

### LINKED ISSUE #66: Create metapackage: Arcus.Scripting
Create metapackage: Arcus.Scripting

Create 1 package which imports all om the Arcus.Scripting.* modules.

stijnmoreels:
Creates a metapackage `Arcus.Scripting.All` which includes all the Arcus PS modules.

Closes #66


stijnmoreels:
Uses `Arcus.Scripting.Storage` from #125 but since only syntax will be checked, this is not a problem here.

mbraekman:
Any reason this is called `Arcus.Scripting.All` and not just `Arcus.Scripting`?

stijnmoreels:
> Any reason this is called `Arcus.Scripting.All` and not just `Arcus.Scripting`?

Ah, consistency with our other Arcus libraries and also something you see in other package groups. Like the `Microsoft.AspNetCore.All` package.
Seems like a good approach no? It explicitly shows that it contains everything on the topic.

mbraekman:
> > Any reason this is called `Arcus.Scripting.All` and not just `Arcus.Scripting`?
> 
> Ah, consistency with our other Arcus libraries and also something you see in other package groups. Like the `Microsoft.AspNetCore.All` package.
> Seems like a good approach no? It explicitly shows that it contains everything on the topic.

Aha, yep if you tell it like that, seems like a good approach indeed 👍

mbraekman (approved):

PR #133, opened 2021-01-29
Testing - add backdoor testing for LogicApps integration tests

### LINKED ISSUE #107: Make sure that the PS Pester tests include backdoor testing
Some backdoor parameter assertion can be added here so we can verify that the correct values are send to these mocked functions. (Like `State = Enabled`, `ResourceGrouop = 'ignored-resource-group'`)

_Originally posted by @stijnmoreels in https://github.com/arcus-azure/arcus.scripting/pull/102#discussion_r526607162_

stijnmoreels:
Adds some backdoor testing for test mocks (technically they were stubs, but with the backdoor testing they became mocks 😉).

Closes #107

### DIFF
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.LogicApps.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.LogicApps.tests.ps1
@@ -238,9 +238,12 @@ Describe "Arcus" {
             }
             It "Disable logic app when stopType = Immediate" {
                 # Arrange
-               $resourceGroupName = "my-resource-group"
+               $resourceGroup = "my-resource-group"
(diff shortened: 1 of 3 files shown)


mbraekman (commented):
Was the renaming of variables "resourceGroupName" to "resourceGroup" done on purpose to avoid overlapping internal resourceGroupName and test-variables? Or is this because I was late in the review-process.

stijnmoreels:
> Was the renaming of variables "resourceGroupName" to "resourceGroup" done on purpose to avoid overlapping internal resourceGroupName and test-variables? Or is this because I was late in the review-process.

Yeah was bc of overlapping. Case insensitive variables so we need them to be different variables to be sure that we assert the right ones.

mbraekman (approved):

PR #156, opened 2021-04-01
chore: streamline parameter order and guards

stijnmoreels:
Use the same approach when describing parameters and verifying their values.


mbraekman (approved):
Nice clean-up 👍

Issue #166, opened 2021-04-23
Get-AzKeyVaultAccessPolicies does not return a consistent result


fgheysels:
**Describe the bug**
When Get-AzKeyVaultAccessPolicies is run against a KeyVault that does not (yet) exists, the return-type is not the same as when it is run against a KeyVault that does exist.

**To Reproduce**
Run the Get-AzKeyVaultAccessPolicies script with parameters that reference a non existing KeyVault.

**Expected behavior**
The return value of this script must be an object with the same structure as when this script is run against a KV which does exist.

**Additional context**
This script is used in an Azure Devops pipeline.  We execute this script just before deploying an Azure KeyVault resource to retrieve the existing keyvault access policies of that KeyVault.  When (re)deploying the KeyVault resource we can then restore the existing access policies.
However, when you run the pipeline for the very first time, the KV obviously doesn't exist.  When you pass in the result of this script to the ARM template for deploying the KV, you then get an error.

mbraekman:
I should have a look at how we handled this in the past, but would indeed make more sense to return an empty _list_ instead.
Thanks for the input.

PR #167, opened 2021-04-23
feat: Return an object with an empty list property when KV does not exist

fgheysels:
When the Get-AzKeyVaultAccesspolicies is run against a non existing KV, make sure that the returned object contains a list property which is empty.
By doing so, the return type is consistent with what we get back when the KV does exist.

#166

### DIFF
--- a/src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
+++ b/src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
@@ -15,11 +15,11 @@ else
 	$keyVault = Get-AzKeyVault -VaultName $keyVaultName -ResourceGroupName $resourceGroupName
 }
 
-if($keyVault)
-{
-	$armAccessPolicies = @()
-	$keyVaultAccessPolicies = $keyVault.accessPolicies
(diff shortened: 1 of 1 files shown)


src/Arcus.Scripting.KeyVault/Scripts/Get-AzKeyVaultAccessPolicies.ps1
```diff
@@ -15,11 +15,11 @@ else
 	$keyVault = Get-AzKeyVault -VaultName $keyVaultName -ResourceGroupName $resourceGroupName
 }
 
-if($keyVault)
-{
-	$armAccessPolicies = @()
-	$keyVaultAccessPolicies = $keyVault.accessPolicies
+$armAccessPolicies = @()
+$keyVaultAccessPolicies = $keyVault.accessPolicies
```
mbraekman, line 19: Any specific reason why you're setting the $keyVaultAccessPolicies-variable outside of the if-statement?
This will only be used within that statement and will remain empty when the $keyVault-variable remains null.

fgheysels: No, should indeed be put inside the if.  I'll change that.

mbraekman (approved):

Issue #174, opened 2021-05-27
Use real Azure resources in our integration tests


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
In the very beginning, we only added scripts to the repo that were already tested in another repo. But this changes and so we are having some blind spots if it comes to the security and stability of our scripts.

**Describe the solution you'd like**
Use where we can real Azure resources (within the price limit) to test our current set of scripts.

stijnmoreels:
Almost all things are (integration) tested. I see the following still open:
* Azure API Management (will not do bc it costs too much)
* Data Factory (don't really sure what's needed, or what it costs)
* Logic Apps (don't really sure what's needed)

Some things to be investigated.

stijnmoreels:
So, we have already integration tests for Logic Apps, API Management is too expensive, and our Data Factory scripts are actually only a single one with a single function; so maybe also not the costs or effort worth?

pim-simons:
> So, we have already integration tests for Logic Apps, API Management is too expensive, and our Data Factory scripts are actually only a single one with a single function; so maybe also not the costs or effort worth?

Only thing missing would indeed be Data Factory. From what I gather running costs would be low, but it would take some effort to build this as you would also need to create an actual flow within Data Factory. To me it seems like a bit too much effort for a single script and we can close this issue. @stijnmoreels  what do you think?

stijnmoreels:
Yeah, @pim-simons , you're right. Looking at the code, the most important additions is exception handling, which is already tested in our unit tests. The script itself uses solely Azure PS commands, so seems indeed that this would be too big an effort.
We could come back to it if our Data Factory module is expanding or problems arise from people. 👍 

Closing this.

PR #176, opened 2021-06-04
test: az bob storage with real az resources

stijnmoreels:
Adds integration tests for Azure Blob storage script that uses an actual Azure Blob container.

Relates to #174


mbraekman (commented):
Only 2 remarks:
- Can you update the documentation to reflect the new and improved output from the script.
- Nitpicking, but I would rename the new field that has been added to the appsettings-file as such:
  `
"StorageAccount": {      "Name": "#{Arcus.StorageAccount.Name}#"     }
`

stijnmoreels:
> * Nitpicking, but I would rename the new field that has been added to the appsettings-file as such:
>   `"StorageAccount": {      "Name": "#{Arcus.StorageAccount.Name}#"     }`

About that.. I was following the namespace here (as in `...Storage.Blob`, `...Storage.Table`) because I imagine there's going to be things added storage-related. And with that, they would all be under:

```json
"Storage": {
    "StorageAccountName": "",
    "Table": { }
    "Blob": { }
}
```

What we may could do is go a level deeper?


```json
"Storage": {
    "StorageAccount":  { "Name": "" },
    "Table": { }
    "Blob": { }
}
```

mbraekman (approved):

PR #187, opened 2021-06-30
test: use real az resources for table storage integration tests

### LINKED ISSUE #174: Use real Azure resources in our integration tests

stijnmoreels:
Use real-life Azure resources for the integration tests of the Azure Table storage PS module.

Relates to #174


mbraekman (approved):
LGTM

PR #196, opened 2021-07-16
fix: prerelease nuget generation

stijnmoreels:
Trouble when building prereleases NuGet packages. Seems like there's a problem with the tool-script we're using.
Including exported DSC resources causing the problem, but since we're not using that, I commented out and documented.


stijnmoreels:
If we have time left in the future, we can look into it further, for now, it seems enough.

mbraekman (approved):
Ok for now, but we'll need to keep this in mind when release an actual version.

PR #198, opened 2021-07-16
Fix/inject arm on linux

### LINKED ISSUE #197: Replace \r\n does not work on linux agent
**Describe the bug**
Replace \r\n does not work on linux agent.
Content injections works fine but I still have the new lines therefore the ARM template is invalid

**To Reproduce**
Run this one on a linux agent
- task: PowerShell@2
  inputs:
    targetType: 'inline'
    script: 'Inject-ArmContent -Path "$(System.ArtifactsDirectory)/dashboards"'
    displayName: Inject KQL queries 

**Expected behavior**
A valid ARM template

**Additional context**
Replace this line
https://github.com/arcus-azure/arcus.scripting/blob/6bc6e1e78080948c37dcd79886887cbd61541ef5/src/Arcus.Scripting.ARM/Scripts/Inject-ArmContent.ps1#L64

with the following code
				$newString = $newString -replace "`r", ""
				if ( [environment]::OSVersion.VersionString -like "*Windows*") {
					$newString = $newString -replace "`n", "\r\n"
				} else {
					$newString = $newString -replace "`n", "\n"
				}

MassimoC:
This PR should close #197  (partially tested)

### DIFF
--- a/src/Arcus.Scripting.ARM/Scripts/Inject-ArmContent.ps1
+++ b/src/Arcus.Scripting.ARM/Scripts/Inject-ArmContent.ps1
@@ -61,7 +61,12 @@ function InjectFile {
                 Write-Host "`t Replacing special characters"
 
                 # Replace newline characters with literal equivalents
-                $newString = $newString -replace "`r`n", "\r\n"
+				$newString = $newString -replace "`r", ""
+				if ( [environment]::OSVersion.VersionString -like "*Windows*") {
+					$newString = $newString -replace "`n", "\r\n"
+				} else {
+					$newString = $newString -replace "`n", "\n"
+				}
 
                 # Replace tabs with spaces
                 $newString = $newString -replace "`t", "    "
--- a/src/Arcus.Scripting.ARM/Scripts/Inject-ArmContent.ps1
+++ b/src/Arcus.Scripting.ARM/Scripts/Inject-ArmContent.ps1
@@ -61,12 +61,12 @@ function InjectFile {
                 Write-Host "`t Replacing special characters"
 
                 # Replace newline characters with literal equivalents
-				$newString = $newString -replace "`r", ""
-				if ( [environment]::OSVersion.VersionString -like "*Windows*") {
-					$newString = $newString -replace "`n", "\r\n"
-				} else {
-					$newString = $newString -replace "`n", "\n"
-				}
+                $newString = $newString -replace "`r", ""
+                if ( [environment]::OSVersion.VersionString -like "*Windows*") {
+                    $newString = $newString -replace "`n", "\r\n"
+                } else {
+                    $newString = $newString -replace "`n", "\n"
+                }
 
                 # Replace tabs with spaces
                 $newString = $newString -replace "`t", "    "


MassimoC:
before the fix
image

with a local version of the fix
image

tomkerkhove (approved):
LGTM, thanks!

mbraekman (approved):
Thanks for this!

stijnmoreels:
Should we make use of the `[System.Environment]::NewLine` here?

MassimoC:
> Should we make use of the `[System.Environment]::NewLine` here?
would it work on linux?

stijnmoreels:
> > Should we make use of the `[System.Environment]::NewLine` here?
> 
> would it work on linux?

Well, it was designed to have a cross-platform approach:
https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline

mbraekman:
> > > Should we make use of the `[System.Environment]::NewLine` here?
> > 
> > 
> > would it work on linux?
> 
> Well, it was designed to have a cross-platform approach:
> https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline

Once again, good thinking!
Would prefer using this approach then instead of the hardcoded characters.

MassimoC:
ok, deal done. I will do the changes and the tests and get back to you

stijnmoreels:
Thanks @MassimoC for your contribution here!

MassimoC:
guys I am trying the following but it replaces the new line with another new line so it doesn't work.
$newString = $newString -replace "`n", ([System.Environment]::NewLine).ToString()

what we need is to replace the new line with a text representation. I am a super rookie here, how do u fix this?

mbraekman:
> guys I am trying the following but it replaces the new line with another new line so it doesn't work.
> $newString = $newString -replace "`n", ([System.Environment]::NewLine).ToString()
> 
> what we need is to replace the new line with a text representation. I am a super rookie here, how do u fix this?

You're not getting the literal string when using `[System.Environment]::NewLine`, but the actual CR/LF, while in this case you'll want to have the literal string.
This is something you can do like this:
```powershell
[System.Environment]::NewLine -replace "`r`n", "\r\n"
```

Which would have the full piece of code look like this:
```powershell
$newString = $newString -replace "`n", $([System.Environment]::NewLine -replace "`r`n", "\r\n")
```

MassimoC:
yeah...what I wanted to say is that, at the end of the day, I do not see a lot of benefits

stijnmoreels:
> yeah...what I wanted to say is that, at the end of the day, I do not see a lot of benefits

The biggest benefit here is that e don't rely on OS information but that's cross-platform by default.

MassimoC:
So I think I misunderstood the comment of Maxim. I thought he wanted to get rid of the hardcoded characters but the goal is to remove the If statement

mbraekman:
> So I think I misunderstood the comment of Maxim. I thought he wanted to get rid of the hardcoded characters but the goal is to remove the If statement

As Stijn indicates, the idea here is indeed to remove the OS-specific dependency.
We'll still need the actual replace-command, as we need to have the literal characters inserted.

stijnmoreels:
@MassimoC do you have enough info to get started? Or do you need some guidance?

PR #199, opened 2021-07-19
Improved usage documentation on Save-AzDevOpsBuild

fgheysels:
Added an example of how to use the Save-AzDevOpsBuild function in an Azure DevOps pipeline.


mbraekman (approved):
Useful addition indeed.
Thanks!

PR #207, opened 2021-08-09
docs: Provide new website template

Mcalewaert:
Relates to https://github.com/arcus-azure/arcus.github.template


tomkerkhove (changes requested):
Can we rename every page's title from `Scripts related to interacting with Azure API Management` to `Azure API Management` please?

Mcalewaert:
@tomkerkhove Done

mbraekman:
Small remark: can we indent the "Azure Blob Storage", "Azure File Share" and "Azure Table Storage" pages to become a sub-page of "Azure Storage"?

Mcalewaert:
@mbraekman Was a bit harder than expected but storage files are now sitting inside an extra directory and I've added redirect rules for the old urls.

tomkerkhove (approved):

PR #211, opened 2021-08-13
test: set arm outputs to az devops variable group (🚧)

stijnmoreels:
Adds integration tests for the PS script that sets the ARM outputs as variables in Azure DevOps variable groups.

Relates to #174

### DIFF
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
@@ -20,7 +20,7 @@ InModuleScope Arcus.Scripting.DevOps {
                 try {
                     # Act
                     Save-AzDevOpsBuild -ProjectId $projectId -BuildId $buildId
-
+        
                     # Assert
                     $getResponse = Invoke-WebRequest -Uri $requestUri -Method Get -Headers $headers
                     $json = ConvertFrom-Json $getResponse.Content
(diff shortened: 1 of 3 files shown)


src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
```diff
@@ -20,7 +20,7 @@ InModuleScope Arcus.Scripting.DevOps {
                 try {
                     # Act
                     Save-AzDevOpsBuild -ProjectId $projectId -BuildId $buildId
-
+
```
fgheysels, line 23:
```suggestion

```
What nitpickers we are ... :)

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
```diff
@@ -33,5 +33,40 @@ InModuleScope Arcus.Scripting.DevOps {
...
+                    # Act
+                    Set-AzDevOpsArmOutputsToVariableGroup -VariableGroupName $existingVariableGroup
+
+                    # Assert
+                    $getVariableGroupUrl= $projectUri + $project + "/_apis/distributedtask/variablegroups?api-version=" + $apiVersion + "&groupName=" + $existingVariableGroup
+                    $variableGroup = Invoke-RestMethod -Uri $getVariableGroupUrl -Headers $headers -Verbose
+                    $variableGroup = $variableGroup.value[0]
+                    $variableGroup.variables.$variableName | Should -Not -Be $null
+                    $variableGroup.variables.$variableName.value | Should -Be "@{value=$variableValue}"
+                #} finally {
```
fgheysels, line 61: Why is the try/finally construct commented out ?

fgheysels (approved):
Looks OK to me.

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
stijnmoreels: Because I'm still working on it. This PR is not ready for a review.

stijnmoreels:
I have a problem with giving access rights to the build user. Normally only two users are to have write permissions on the Azure DevOps variable group, but that doesn't seem to be enough. Could it be that the request comes from a GitHub repo instead of a Azure DevOps repo? Maybe some additional user requires permissions, only not sure who that would be.

Currently, the tests fail with `Forbidden`.

pim-simons:
Ran into the same issue while create an integration test in this PR https://github.com/arcus-azure/arcus.scripting/pull/252.
While updating the variable group the test receives back an HTTP 403.

It turns out that when you start the pipeline manually it does work, so there seems to be an issue with the user used by the automatically triggered runs.

stijnmoreels:
Close in favor of #252

PR #217, opened 2021-08-20
Changes to support Semantic Versioning on DB changes

fgheysels:
Added some changes to support semantic versioning for the database migrations.
This means that, instead of having DB versions like, 1, 2, 3, .... , 854231, you can now also use semanic versioning.  When a DB migration only consist of adding some records to a certain table for instance, you can increase the patch-part of the version number (for instance 1.0.0 -> 1.0.1).    When new tables are added, increase the major version number part, etc....

This change also means that the filenames of the scripts that contain the DB changes, should no longer be named {number}_{description}.sql (f.i. 1_initialversion.sql) but {major.minor.patch}_{description}.sql  (f.i. 1.0.0_initialversion.sql)

Additional:  I've removed the hard-coded `.database.windows.net` in the ServerInstance part because I feel that we should not have a hard dependency on this.  I have had a situation where I wanted to use this script for a SQL Server DB which was not hosted in Azure f.i.   I just think you need to provide the complete ServerInstance as a parameter.


fgheysels:
Note:  I think we should also deliver 'migration scripts' for modifying existing filenames and existing 'DatabaseVersion' tables.   I just don't know how we should package or deliver those .

stijnmoreels (changes requested):
Great idea to improve how we handle migration versions! 🥇 

Some remarks:
- [x] The current changes are rather breaking, we should consider making this backwards compatible so existing migration versions still work (the current unit tests can help with this)
- [ ] The (preview) feature documentation should reflect this change
- [ ] We are still working on an integration test for the SQL migration-stuff. This can be a good opportunity to add those.

Is there something you can take up @fgheysels ? New unit tests and integration tests, feature docs...
On tests I'm thinking that we should also make sure that we handle migration versions correctly:
- [x] Verify if lower migration is ignored (lower patch, lower, minor, lower major)
- [x] Verify if invalid migration version is ignored (not a number, not three numbers, negative numbers...)
- [ ] Verify if a higher migration is run (higher patch, higher minor, higher major)
- [x] Verify if same migration version are ignored

Some of these can become unit tests where we stub-out the file names, but we should also have some integration tests to verify the major paths.

fgheysels:
> * [ ]  The current changes are rather breaking, we should consider making this backwards compatible so existing migration versions still work (the current unit tests can help with this)

The changes are indeed breaking; that's why I also commented that it would be necessary to provide some kind of migration path (script to rename files, script to modify current DatabaseVersion table).
I'm just not sure what the best option would be:

- adding those scripts to the repository, and mention this in the documentation + document how those scripts must be used. 
- Make sure that running the script performs some checks first, and auto-migrates existing situations ?  (Don't know if this is worth the effort)

I can start creating those migration scripts and have a look at updating the documentation as well.

stijnmoreels:
> > * [ ]   The current changes are rather breaking, we should consider making this backwards compatible so existing migration versions still work (the current unit tests can help with this)
> 
> The changes are indeed breaking; that's why I also commented that it would be necessary to provide some kind of migration path (script to rename files, script to modify current DatabaseVersion table).
> I'm just not sure what the best option would be:
> 
> * adding those scripts to the repository, and mention this in the documentation + document how those scripts must be used.
> * Make sure that running the script performs some checks first, and auto-migrates existing situations ?  (Don't know if this is worth the effort)
 
I think it's best to make this either backwards compatible (script can handle both type of versioning) or we migrate automatically with this new versions, because that would be the best user-experience and I'm also thinking that not a lot of people will take the effort of reading the documentation to run some mandatory scripts. 🙄 

> I can start creating those migration scripts and have a look at updating the documentation as well.

Great!

fgheysels:
> > > * [ ]    The current changes are rather breaking, we should consider making this backwards compatible so existing migration versions still work (the current unit tests can help with this)
> > 
> > 
> > The changes are indeed breaking; that's why I also commented that it would be necessary to provide some kind of migration path (script to rename files, script to modify current DatabaseVersion table).
> > I'm just not sure what the best option would be:
> > 
> > * adding those scripts to the repository, and mention this in the documentation + document how those scripts must be used.
> > * Make sure that running the script performs some checks first, and auto-migrates existing situations ?  (Don't know if this is worth the effort)
> 
> I think it's best to make this either backwards compatible (script can handle both type of versioning) or we migrate automatically with this new versions, because that would be the best user-experience and I'm also thinking that not a lot of people will take the effort of reading the documentation to run some mandatory scripts. 🙄

I agree that this gives the best user experience, but we should also balance cost/effort and what we gain with it.   I don't know f.i. if there're already lots of people using this.   If only 10 people are using it right now already, this might not be worth the effort

gverstraete:
The more I look at this PR the more I see you want to become kind of a copy of the Migration scripts of EF Core... Not sure if we are missing the ball of this script here...

fgheysels:
> The more I look at this PR the more I see you want to become kind of a copy of the Migration scripts of EF Core... Not sure if we are missing the ball of this script here...

Not sure what you're refering to, or what you exactly mean ?
The only thing that has changed, is the fact that these changes now make sure that semantic versioning can be applied to your DB.  (As far as I remember, EF uses dates for the version numbers ? )

src/Arcus.Scripting.Sql/Scripts/Invoke-AzSqlDatabaseMigration.ps1
```diff
@@ -45,110 +56,149 @@ function Get-SqlScriptFileText([string] $scriptPath, [string] $fileName)
...
 }
 
 #Group params needed to connect to database for ease of use
 $params = Create-DbParams $DatabaseName $ServerName $UserName $Password
 
-#Get all tables on the database
-$query = 'SELECT TABLE_NAME FROM information_schema.tables'
-$tables = Execute-DbCommandWithResult $params $query
+# Create the DatabaseVersion table if it doesn't exist yet
+$createDatabaseVersionTable = "IF NOT EXISTS ( SELECT * FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 'DatabaseVersion' AND TABLE_SCHEMA = '$DatabaseSchema' ) " +
```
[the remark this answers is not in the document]
fgheysels: `INFORMATION_SCHEMA` is a schema that is available in all SQL Server version.
`TABLES` is a view in that schema that gives you information regarding the tables.  I think this view offers a layer of abstraction and hides the internal ways to retrieve meta-information of a database that is specific per SQL Server version.

https://docs.microsoft.com/en-us/sql/relational-databases/system-information-schema-views/system-information-schema-views-transact-sql?view=sql-server-2017

https://docs.microsoft.com/en-us/sql/relational-databases/system-information-schema-views/tables-transact-sql?view=sql-server-2017

fgheysels:
I've made some changes to the script to support the older versions.
In other words, this script is now backwards compatible with previous versions:

- When it runs, the script now also checks if the DatabaseVersion table is still in the old format; if so, it upgrades the table
- When the script encounters migration files that follow the old naming convention ({versionnr}_{description}), it converts the old versionnumber to {versionnr}.0.0 and writes a warning that you should change the filenames.

For instance:
image

I will need some help though for unit-testing the script, as I see now that the tests only verify if certain methods have been called.
It would be nice if we could test the automatic upgrade of the databaseversion table (although I think this is rather an integration test instead of a unit-test).

fgheysels:
Provide some additional unit-tests:
- test DatabaseVersion
- verify existing unit-tests / rework

stijnmoreels:
> Provide some additional unit-tests:
> 
> * test DatabaseVersion
> * verify existing unit-tests / rework

In order to test the `DatabaseVersion`, we should include the class in the root module, and not the script.

src/Arcus.Scripting.Sql/Scripts/Invoke-AzSqlDatabaseMigration.ps1
```diff
@@ -1,25 +1,36 @@
...
-    [Parameter(Mandatory=$true)][string] $Password = $(throw "Please provide the Password of the SQL Database"),
+    [Parameter(Mandatory=$true)][string] $UserName = $(throw "Please provide the user name of the user that must be used to perform the update"),
+    [Parameter(Mandatory=$true)][string] $Password = $(throw "Please provide the password of the user that must be used to perform the update"),
     [Parameter(Mandatory=$false)][string] $ScriptsFolder = "$PSScriptRoot/sqlScripts",
     [Parameter(Mandatory=$false)][string] $ScriptsFileFilter = "*.sql",
     [Parameter(Mandatory=$false)][string] $DatabaseSchema = "dbo"
 )
 
+#Import needed for Azure Powershell Release Step
+Import-Module SqlServer
```
stijnmoreels, line 12: Why is this here again? We can added it to the `RequiredModules` in the `.psd1` file, but it should not be here I suppose.

fgheysels: Removed it.

stijnmoreels:
Provided an example of how the integration tests can be written. We should make sure that we have a least some tests to verify some critical points. But, let's not make this a super-blocking factor and add some additional tests in a separate PR.

docs/preview/02-Features/powershell/azure-sql.md
```diff
@@ -55,27 +55,18 @@ PS> RunDatabaseScript -ServerName "my-server-name" -DatabaseName "my-database-na
...
-    [CurrentVersionNumber] INT NOT NULL,
-    [MigrationDescription] [nvarchar](256) NOT NULL,
-    CONSTRAINT [PKDatabaseVersion] PRIMARY KEY CLUSTERED
-    ( 	
-        [CurrentVersionNumber] ASC
-    )
-    WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON)
-)
-```
+1. In the location where you want to run the script add the folder `sqlScripts`.
```
stijnmoreels, line 58: This is actually configurable, maybe that should be mentioned here. ("As an example, we use ..."). or something like that.

docs/preview/02-Features/powershell/azure-sql.md
```diff
@@ -55,27 +55,18 @@ PS> RunDatabaseScript -ServerName "my-server-name" -DatabaseName "my-database-na
...
-Arcus_001_AddIndexes.sql
+`1.0.0_Baseline.sql`
+`1.1.0_AddIndexes.sql`
+`1.1.1_PopulateCodetables.sql`
 
 When a new migration comes along, just create the new SQL script with a version number one number higher than the previous one.
+
+### Compatibility
+
+Semantic versioning of database-migrations is supported since version [XXX].  Existing migration scripts that follow the old naming convention will be recognized and will be given this version-number: `[VersionNumber].0.0`.
```
stijnmoreels, line 72:
```suggestion
Semantic versioning of database-migrations is supported since version v0.5.  Existing migration scripts that follow the old naming convention will be recognized and will be given this version-number: `[VersionNumber].0.0`.
```

Next minor version is v0.5.

src/Arcus.Scripting.Sql/Arcus.Scripting.Sql.psm1
```diff
@@ -1,3 +1,73 @@
+Class DatabaseVersion : System.IComparable
```
stijnmoreels, line 1: Wondering if we should also make sure that this is formatted the same way as our other files 

```powershell
class DatabaseVersion : System.IComparable {
```

stijnmoreels (approved):
I think this is almost done. Amazing work done here!

src/Arcus.Scripting.Sql/Arcus.Scripting.Sql.psm1
fgheysels: You mean we should format it with the opening curly brace on the same line as the if statement or function/class declaration ?
I've modified the `DatabaseVersion` class now so that it is formatted like that.

fgheysels: Then, the same should be done for the migration-script itself imho.

However, I can see that the formatting we're using is not consistent throughout the codebase.

docs/preview/02-Features/powershell/azure-sql.md
fgheysels: Made some changes to make this more clear

src/Arcus.Scripting.Sql/Arcus.Scripting.Sql.psm1
fgheysels: Changed the migrate-script as well to be conform with Powershell coding convention.

PR #237, opened 2021-11-22
fix: fix errors in documentation

fgheysels:
Edit code samples in SQL module documentation


stijnmoreels (approved):
👌 super. Thx!

stijnmoreels:
I think this is already released, so you can update the documentation in the released docs too, if you want.

fgheysels:
> I think this is already released, so you can update the documentation in the released docs too, if you want.

I'll see  how it can be done, and update this PR.

fgheysels:
> > I think this is already released, so you can update the documentation in the released docs too, if you want.
> 
> I'll see how it can be done, and update this PR.

Should be done.

stijnmoreels (approved):
Great 👍

PR #240, opened 2021-11-23
docs: add versioned module installations

stijnmoreels:
Add `-MaximumVersion` to `Install-Module` commands to be in line with the other Arcus docs, and not to wrongly guide consumers to the wrong version.

### DIFF
--- a/docs/preview/01-index.md
+++ b/docs/preview/01-index.md
@@ -5,8 +5,6 @@ slug: /
 sidebar_label: Welcome
 ---
 
-![PowerShell Gallery Version](https://img.shields.io/powershellgallery/v/Arcus.Scripting.All)
-
 # Installation
 
 The Arcus.Scripting modules can be found on the PowerShell gallery.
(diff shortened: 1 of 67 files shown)


mbraekman:
Would you use `-MinimumVersion` or `-RequiredVersion`?
Since `-MininumVersion` might also trigger the download of a newer available version, which might not be in line with that specific documentation-page.

Otherwise: 👍

stijnmoreels:
> Would you use `-MinimumVersion` or `-RequiredVersion`? Since `-MininumVersion` might also trigger the download of a newer available version, which might not be in line with that specific documentation-page.
> 
> Otherwise: 👍

Oh, shoot, you're right. I mend, `-MaximumVersion.` 😅 .

stijnmoreels:
But maybe, `-RequiredVersion` is even better. 🤔 ...

mbraekman:
> But maybe, `-RequiredVersion` is even better. 🤔 ...

-RequiredVersion is probably the only way you're 100% sure that the downloaded version will match what has been specified in the docs, so I would go for that option 😄

mbraekman (approved):

PR #244, opened 2021-12-03
fix: arm inject newline chars on windows

### LINKED ISSUE #242: Inject-ArmContent: replacing newline characters doesn't work as expected anymore
**Describe the bug**
For Build in Azure DevOps, I use the Inject-ArmContent of Arcus.Scripting.ARM to inject an xml policy inside a json arm template for an API.

`"policyContent": "${ FileToInject='policies\\inventorymgmt-dst-swire-base-policy.xml', EscapeJson, ReplaceSpecialChars }$"`

With a file like that:
```
<policies>
  <inbound>
    <base />
```
 
It works fine for months and amongst other things replaces the carriage returns by the characters \r\n :
 
`"policyContent": "<policies>\r\n  <inbound>\r\n    <base />\r\n   \r\n ` 
 
But now, without any change in my files, this does not work anymore and generates the following on multiple lines (and deployment fails…☹):
 
```
        "policyContent": "<policies>
\r\n  <inbound>
\r\n    <base />
\r\n   
```

stijnmoreels:
Reverts original newline replacement on Windows that was removed in https://github.com/arcus-azure/arcus.scripting/commit/f41369aed4ac1d4d1e8b445ef4ba85f03a1de2c2.
Harden the integration tests so we check for new line characters.

Closes #242


src/Arcus.Scripting.Tests.Integration/Files/arm-template-object-value (linux).json
```diff
@@ -0,0 +1,3 @@
+﻿{
+   "test": "this is a test value"
```
fgheysels, line 2: Don't we need testfiles that contain newline chars ?

stijnmoreels: This test file contains the injected value for Linux. The end-result will contain these newline chars (see assertion). That's also the fix of this PR: multiline file to newline chars.

fgheysels (approved):
So the change has just been reverted, and the previous version has been set back ?

stijnmoreels:
> So the change has just been reverted, and the previous version has been set back ?

Yeah, for windows this has been reverted to the same thing we had before v0.5.

fgheysels:
> > So the change has just been reverted, and the previous version has been set back ?
> 
> Yeah, for windows this has been reverted to the same thing we had before v0.5.

So we need a followup- PR then  or not ?

fgheysels (approved):

stijnmoreels:
> > > So the change has just been reverted, and the previous version has been set back ?
> > 
> > 
> > Yeah, for windows this has been reverted to the same thing we had before v0.5.
> 
> So we need a followup- PR then or not ?

No, no need, I reverted these changes in this PR.

pim-simons (approved):
Finally had some time to take a look, looks good to me 👍🏻

PR #246, opened 2021-12-15
fix: catch 'activate' statement

stijnmoreels:
Let's see if this works.

### DIFF
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
@@ -94,12 +94,15 @@ InModuleScope Arcus.Scripting.Sql {
 
             & $PSScriptRoot\Connect-AzAccountFromConfig.ps1 -config $config
 
-            # Try to open a connection to the SQL database, so that the
-            # Azure Database that can be paused, is starting up.  This should
-            # avoid having timeout errors during the test themselves.
-            # We don't care if an exception is thrown; we just want to 'activate' the Azure SQL database
-            Write-Host "Execute dummy SQL statement to make sure the Azure SQL DB is resumed."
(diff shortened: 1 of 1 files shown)


src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
```diff
@@ -94,12 +94,15 @@ InModuleScope Arcus.Scripting.Sql {
...
-            Invoke-Sqlcmd @params -Query "SELECT TOP 1 TABLE_NAME FROM INFORMATION_SCHEMA.TABLES" -ConnectionTimeout 60 -Verbose -ErrorAction SilentlyContinue
+            # Try to open a connection to the SQL database, 
+            # so that the Azure Database that can be paused, is starting up.  
+            # This should avoid having timeout errors during the test themselves.
+            try {
+                Write-Host "Execute dummy SQL statement to make sure the Azure SQL DB is resumed."
+                Invoke-Sqlcmd @params -Query "SELECT TOP 1 TABLE_NAME FROM INFORMATION_SCHEMA.TABLES" -ConnectionTimeout 60 -Verbose -ErrorAction SilentlyContinue
+            } catch {
+                # We don't care if an exception is thrown; we just want to 'activate' the Azure SQL database.
+            }
```
fgheysels, line 105: Oh, I once removed that try / catch because of the 'SilentlyContinue' error-action, but it seems it is necessary anyway ?

fgheysels (approved):

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
stijnmoreels: Seemed like it was necessary in the windows-image integration tests. The linux integration tests don't need this. Maybe something wrong with the SQL statement bc it gave syntax errors; but not sure if we should spent time investing as we don't care of the exception. So, added a catch statement to be sure.

pim-simons (approved):

Issue #251, opened 2021-12-22
Set-AzDevOpsArmOutputs is missing the release in the description of the library when using YAML pipelines


pim-simons:
**Describe the bug**
When using the Set-AzDevOpsArmOutputs script the description of the library is updated but the releasename is missing:
![image](https://user-images.githubusercontent.com/32359437/147056028-b63feb85-5a46-485b-bd12-de87a54f2cef.png)
This only seems to happen when using YAML pipelines.

**To Reproduce**
Use the Set-AzDevOpsArmOutputs script in a YAML pipeline. 

**Expected behavior**
The library description to have the pipeline name. 

**Additional context**
 - Version: latest version of the script

PR #252, opened 2021-12-22
fix: make sure the variable group description uses a releasename when using YAML pipelines

pim-simons:
When using YAML pipelines the Release.ReleaseName is empty, if this is the case use Build.DefinitionName and Build.BuildNumber and use this as the releasename in the variable group description.

Closes https://github.com/arcus-azure/arcus.scripting/issues/251

### DIFF
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
@@ -32,6 +32,23 @@ InModuleScope Arcus.Scripting.DevOps {
                     $patchResponse.StatusCode | Should -Be 200
                 }
             }
+            It "Sets the DevOps variable group description with the release name" {
+                # Arrange
+                $variableGroupName = $config.Arcus.DevOps.VariableGroup.Name
+                $env:ArmOutputs = "{ ""$variableGroupName"": [ { ""Name"": ""my-variable"", ""Value"": { ""value"": ""my-value"" } } ] }"
+                $projectId = $env:SYSTEM_TEAMPROJECTID                
+                $collectionUri = $env:SYSTEM_COLLECTIONURI
+                $requestUri = "$collectionUri" + "$projectId/_apis/distributedtask/variablegroups?groupName=/" + $variableGroupName + "?api-version=6.0"
+                $headers = @{ Authorization = "Bearer $env:SYSTEM_ACCESSTOKEN" }
+
+                # Act
+                Set-AzDevOpsArmOutputsToVariableGroup -VariableGroupName $variableGroupName
+
+                # Assert
+                $getResponse = Invoke-WebRequest -Uri $requestUri -Method Get -Headers $headers
+                $json = ConvertFrom-Json $getResponse.Content
+                $json.description | Should -BeLike "*$env:Build_DefinitionName*$env:Build_BuildNumber*"
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


stijnmoreels (changes requested):
Great contribution from a real practical problem! Nice!
We should test this, though. Make sure that we indeed use a different kind of release name in case it's missing.

pim-simons:
> Great contribution from a real practical problem! Nice! We should test this, though. Make sure that we indeed use a different kind of release name in case it's missing.

Will see if I can add some tests for this 👍🏻

pim-simons:
> > Great contribution from a real practical problem! Nice! We should test this, though. Make sure that we indeed use a different kind of release name in case it's missing.
> 
> Will see if I can add some tests for this 👍🏻

Added tests, see [`4ae0b8c` (#252)](https://github.com/arcus-azure/arcus.scripting/pull/252/commits/4ae0b8cfe67ff5b07102238723a2c3fd875a0189)

stijnmoreels (approved):
Ok, great, maybe we should also test if both are present, to verify which one is getting the advantage, right?

stijnmoreels:
Question: I see that we don't have integration tests for this. I can't remember why that was, do you think that we can write those? Or are we missing something?

pim-simons:
> Ok, great, maybe we should also test if both are present, to verify which one is getting the advantage, right?

The classic release pipelines only have the Release.ReleaseName available (https://docs.microsoft.com/en-us/azure/devops/pipelines/release/variables?view=azure-devops&tabs=batch) and YAML pipelines never have these. So its more of a one or the other kind of situation. 

I could easily add the test, but I don't think this situation would happen in the "real world", so not sure if that is something we want to test?

pim-simons:
> Question: I see that we don't have integration tests for this. I can't remember why that was, do you think that we can write those? Or are we missing something?

Will take a look, I think we should be able to use this API to get information about the variable group: https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/variablegroups/get-variable-groups?view=azure-devops-rest-6.1

pim-simons:
> > Question: I see that we don't have integration tests for this. I can't remember why that was, do you think that we can write those? Or are we missing something?
> 
> Will take a look, I think we should be able to use this API to get information about the variable group: [docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/variablegroups/get-variable-groups?view=azure-devops-rest-6.1](https://docs.microsoft.com/en-us/rest/api/azure/devops/distributedtask/variablegroups/get-variable-groups?view=azure-devops-rest-6.1)

I tried added an integration test but ran into the same issue as this PR https://github.com/arcus-azure/arcus.scripting/pull/211
While updating the variable group the test receives back an HTTP 403.

It turns out that when you start the pipeline manually it does work, so there seems to be an issue with the user used by the automatically triggered runs.

For now I have removed (commented out) the integration test, we can probably pick this up further in https://github.com/arcus-azure/arcus.scripting/pull/211

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
```diff
@@ -32,6 +32,23 @@ InModuleScope Arcus.Scripting.DevOps {
                     $patchResponse.StatusCode | Should -Be 200
                 }
             }
+            #It "Sets the DevOps variable group description with the release name" {
```
stijnmoreels, line 35: We can maybe use `-Skip` for this: https://pester-docs.netlify.app/docs/usage/skip

stijnmoreels (approved):
Great! Maybe we can actually remove #211  because we already have the skipped test here, right?

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.DevOps.tests.ps1
pim-simons: Better solution indeed, I changed it to `-Skip` 👍🏻

Issue #249, opened 2021-12-21
Broken preview API Management link


stijnmoreels:
**Describe the bug**
In the past, we used different links, which makes some documentation links dead.
https://scripting.arcus-azure.net/preview/features/powershell/azure-api-management is used in this blog: https://www.codit.eu/blog/announcing-arcus-scripting/?country_sel=be but is not active anymore.

**To Reproduce**
Surf to https://scripting.arcus-azure.net/preview/features/powershell/azure-api-management.

**Expected behavior**
Redirect to API Management features.

**Additional context**
 - Current published docs. v1.4.0.

Issue #250, opened 2021-12-21
Broken Azure Storage documentation link 


stijnmoreels:
**Describe the bug**
In the past, we used different links, which makes some documentation links dead.
https://scripting.arcus-azure.net/features/powershell/azure-storage is used in this blog: https://www.codit.eu/blog/announcing-arcus-scripting/?country_sel=be but is not active anymore.

**To Reproduce**
Surf to https://scripting.arcus-azure.net/features/powershell/azure-storage.

**Expected behavior**
Redirect to Azure Storage features.

**Additional context**
 - Current published docs. v1.4.0.

PR #253, opened 2021-12-22
fix: added redirects for broken links

pim-simons:
Added redirects for the broken API Management and Azure Storage documentation links.

Closes https://github.com/arcus-azure/arcus.scripting/issues/249 and https://github.com/arcus-azure/arcus.scripting/issues/250


stijnmoreels:
Ok, great! I checked the deploy preview and it's indeed fixed now. Good job! 👍

stijnmoreels (approved):

PR #263, opened 2022-01-12
test: drop sql tables each test run

### LINKED ISSUE #261: Delete all SQL tables before each test run to have a clean slate
When cancelling a test run, tables can be left in the SQL server which would cause trouble on the next run.

We should remove all SQL tables before each run so we start with a clean slate.

**Additional context**
Relates to #245

stijnmoreels:
Drops all SQL tables from our test SQL server on each test suite run.

Closes #261

### DIFF
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
@@ -104,19 +104,13 @@ InModuleScope Arcus.Scripting.Sql {
                 # We don't care if an exception is thrown; we just want to 'activate' the Azure SQL database.
             }
 
-            $tableNames = @()
-            try {
-                $tableNames = Run-AzSqlQuery $params "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES"
-            } catch {
-                Write-Warning "Could not select any SQL tables to drop during test setup"
-            }
-            
-            foreach ($tableName in $tableNames.ItemArray) {
+            $tables = Run-AzSqlQuery $params "SELECT TABLE_SCHEMA, TABLE_NAME FROM INFORMATION_SCHEMA.TABLES"
+            foreach ($table in $tables) {
                 try {
-                    Write-Verbose "Drop table $tableName in SQL database"
-                    Drop-AzSqlDatabaseTable $params $tableName
+                    Write-Verbose "Drop table $($table.TABLE_NAME) in SQL database"
+                    Drop-AzSqlDatabaseTable $params $table.TABLE_NAME $table.TABLE_SCHEMA
                 } catch {
-                    Write-Warning "Could not drop table '$($tableName)' due to an exception: $($_.Exception.Message)"
+                    Write-Warning "Could not drop table '$($table.TABLE_NAME)' due to an exception: $($_.Exception.Message)"
                 }
             }
         }
--- a/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
+++ b/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
@@ -34,7 +34,7 @@ function global:Get-AzSqlDatabaseVersion ($params, $schema = "dbo") {
     return $version
 }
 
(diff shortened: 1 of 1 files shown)


stijnmoreels:
I probably has to add `.ItemArray` to this.

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
```diff
@@ -103,6 +103,22 @@ InModuleScope Arcus.Scripting.Sql {
             } catch {
                 # We don't care if an exception is thrown; we just want to 'activate' the Azure SQL database.
             }
+
+            $tableNames = @()
+            try {
+                $tableNames = Run-AzSqlQuery $params "SELECT TABLE_NAME FROM INFORMATION_SCHEMA.TABLES"
+            } catch {
+                Write-Warning "Could not select any SQL tables to drop during test setup"
```
pim-simons, line 111: Do we want to stop processing if the retrieval of tables fails? If we continue if this fails we execute tests based on already existing tables and the tests might fail.
In that case this should be a `throw` right?

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
```diff
@@ -103,6 +103,22 @@ InModuleScope Arcus.Scripting.Sql {
...
+            }
+            
+            foreach ($tableName in $tableNames.ItemArray) {
+                try {
+                    Write-Verbose "Drop table $tableName in SQL database"
+                    Drop-AzSqlDatabaseTable $params $tableName
```
pim-simons, line 117: This is not taking into account specific schemas, it defaults back to `dbo` due to `function global:Drop-AzSqlDatabaseTable ($params, $databaseTable, $schema = "dbo")` on line 19.

Not sure if we have any tables that are in a specific schema other than `dbo` that need to be dropped?

pim-simons (commented):
Looks good, just some small questions 👍🏻

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
stijnmoreels: Hmm, good catch. The information_schema query will get all our tables, but doesn't take into account schemes, indeed. Will need to add that.

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
stijnmoreels: Yeah, this gave problems on Linux images. Still not sure why. I guess, since we run this on both environments, at least one will succeed. But didn't got the time yet to look into why it sometimes fails.

pim-simons: But if it fails, do we want to only write a warning and continue with the script?

stijnmoreels: I'll try to look into this. It's an insurance policy, these changes. The tests itself should handle the teardown. It's a back-up of a back-up, really. But still valid 😉 . I'll try to investigate in this, why it fails.

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
stijnmoreels: Added in https://github.com/arcus-azure/arcus.scripting/pull/263/commits/f1bd166a4bd475a944300c7c18cb9e02654d40b6

stijnmoreels:
Aha! All green!

pim-simons (approved):

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Sql.tests.ps1
stijnmoreels: Failure was a transient Azure SQL failure, so would need to retry that which made the test succeed.

Issue #226, opened 2021-10-07
Provide documentation on various modes for injecting content into an ARM template


tomkerkhove:
Provide documentation on various modes for injecting content into an ARM template and what they do.

For example:
- `InjectAsJsonObject`
- `EscapeJson`
- `ReplaceSpecialChars`

PR #269, opened 2022-02-03
feat: Provide documentation on various modes for injecting content into an ARM template

pim-simons:
Added documentation on the injection instructions into an ARM template and what they do.

Closes https://github.com/arcus-azure/arcus.scripting/issues/226

### DIFF
--- a/docs/preview/02-Features/powershell/arm.md
+++ b/docs/preview/02-Features/powershell/arm.md
@@ -28,7 +28,7 @@ By using this script, you can inject external files inside your ARM template.
 | --------- | --------- | ----------------------------------------------------------------------------------------------- |
 | `Path`    | no        | The file path to the ARM template to inject the external files into  (default: `$PSScriptRoot`) |
 
-**Usage**
+### Usage
 Annotating content to inject:
 
 ```json
@@ -54,7 +54,7 @@ Injecting the content:
 PS> Inject-ArmContent -Path deploy\arm-template.json
 ```
 
-**Injection Instructions**
+### Injection Instructions
 It is possible to supply injection instructions in the injection annotation, this allows you to add specific functionality to the injection. These are the available injection instructions:
 
 | Injection Instruction | Description                                                                                                 |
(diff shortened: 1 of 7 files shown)


docs/preview/02-Features/powershell/arm.md
```diff
@@ -54,5 +54,23 @@ Injecting the content:
...
 
-**Recommendations**
+### Injection Instructions
+It is possible to supply injection instructions in the injection annotation, this allows you to add specific functionality to the injection. These are the available injection instructions:
+
+| Injection Instruction | Description                                                                                                 |
+| --------------------- | ----------------------------------------------------------------------------------------------------------- |
+| `EscapeJson`          | Replace double quotes not preceded by a backslash with escaped quotes                                       |
+| `ReplaceSpecialChars` | Replace newline characters with literal equivalents, tabs with spaces and `"` with `\"`                     |
+| `InjectAsJsonObject`  | Tests if the content is valid JSON and makes sure the content is injected without surrounding double quotes |
```
stijnmoreels, line 64: Maybe we can add this update to the already released version. We could look if these things are already released, and update those releases too, right?

stijnmoreels (approved):
Great stuff! Valuable update, Pim. Was already 'lingering' on our issue list. Great initiative.

docs/preview/02-Features/powershell/arm.md
pim-simons: Since it is not a code change and just documentation I think it should be no problem to add this to the already release version. I haven't done that before though, so I would need some guidance on how to do that.

stijnmoreels: When releasing the feature docs (in any Arcus repo), we run `npm run docusaursus docs:version vx.x.x` (as described [here](https://github.com/arcus-azure/arcus.scripting/tree/master/docs#versioning)). This results in a specific version folder in the `/docs` folder in the root of the repo. So for (example) version `v1.1.0`, will result in a folder `/docs/versioned_docs/version-v1.1.0` where the same structure is available as in the `/docs/preview` folder. 

So, in short, the only difference between preview and published docs, is `/docs/preview` and `/docs/versioned_docs/version-vx.x.x` folders.

I think the ARM injection is already published for quite some time, so it may be available already very early. Looking in the tags of the repo (https://github.com/arcus-azure/arcus.scripting/tree/v0.4.3) can help.

pim-simons: I have included it in the previous versions of the documentation as well, see [`e7a388f` (#269)](https://github.com/arcus-azure/arcus.scripting/pull/269/commits/e7a388f3fbc12314b1332702487168ade2483864)

stijnmoreels (approved):
Perfect! Thx. 👌

Issue #190, opened 2021-07-14
Provide script to purge soft-deleted resources


mbraekman:
**Is your feature request related to a problem? Please describe.**
Currently one cannot purge soft-deleted services via PowerShell, this has to be done via a DELETE operation.
By providing this PS-functionality, one doesn't have to use a tool or script to build the proper request before being able to purge the service.

**Describe the solution you'd like**
Suggested method:
`Remove-AzSoftDeletedService`

Parameters:
- SubscriptionId:
  The Id of the subscription in which the soft-deleted service exists.
- ServiceType:
  The type of service which is to be purged.
  For example: Microsoft.ApiManagement
- Location:
  The region where the soft-deleted service was created.
  For example: WestEurope
- ServiceName:
  The name of the service which is to be purged.
  For example: arcus-dev-we-gateway

**Describe alternatives you've considered**
The alternative would be to use a tool, such as PostMan, to create the DELETE-request to purge the service.

**Additional context**
See Microsoft documentation on how to purge a soft-deleted APIM-instance for example: [here](https://docs.microsoft.com/en-us/azure/api-management/soft-delete#purge-a-soft-deleted-apim-instance)

The same could be created to recover soft-deleted services:
- `Restore-AzSoftDeletedService`

Or to list the soft-deleted services:
- `Get-AzSoftDeletedServices`

mbraekman:
Open to suggestions as to which "domain" this should belong to.

stijnmoreels:
Maybe this belongs in a 'general' domain like we have one for security and devops.

mbraekman:
> Maybe this belongs in a 'general' domain like we have one for security and devops. But great idea!
> 
> I was thinking of `Arcus.Scripting` but that's maybe to general and people won't know directly what's about. Maybe under something called 'management' or 'resource management'.

I like the idea about Arcus.Scripting.Management :)

pim-simons:
@mbraekman what resources are you thinking of to support in this script? Initially just APIM or others as well?

mbraekman:
@pim-simons I'd start with APIM and perhaps all of those resource types which are currently supported by 1 or another script within Arcus.
Could easily be extended later on of course.

pim-simons:
Did a small inventorisation of the resource types we support within Arcus and whether or not they support soft-delete and the current scripting support for purging and recovering.

1. App Service - Soft-delete not available
2. Data Factory - Soft-delete not available
[...]

PR #278, opened 2022-03-10
feat: added apim soft delete scripts

pim-simons:
Added scripts to remove and restore soft-deleleted Azure APIM instances.
Closes https://github.com/arcus-azure/arcus.scripting/issues/190

Note, haven't been able to add integration tests, as this would entail creating, restoring and removing Azure APIM instances which takes about 45 mins, a bit too long to use in the integration test.


pim-simons:
I commented out the unit tests for now, the scripts contain this code which I cannot mock from the unit tests:
``` powershell
$azProfile = [Microsoft.Azure.Commands.Common.Authentication.Abstractions.AzureRmProfileProvider]::Instance.Profile
if (-not $azProfile.Accounts.Count) {
    throw "Ensure you have logged in (Connect-AzAccount) before calling this function"
}
$currentAzureContext = Get-AzContext
$subscriptionId = $currentAzureContext.Subscription.id
$profileClient = New-Object Microsoft.Azure.Commands.ResourceManager.Common.RMProfileClient($azProfile)
$token = $profileClient.AcquireAccessToken($currentAzureContext.Tenant.TenantId)
$authHeader = @{
   'Authorization'='Bearer ' + $token.AccessToken
}
```

Since we are working with REST API's I need the AccessToken to authenticate. And the unit tests fail because there is no `azProfile.Accounts` because the Connect-AzAccount has not been executed for the unit tests.

I used the same code as used in `Get-AzCachedAccessToken.ps1` and unfortunately there are no unit tests for `Get-AzCachedAccessToken.ps1` that I can copy from for this. 

@stijnmoreels any ideas on how to go about this? I see that the integration tests contain a `Connect-AzAccountFromConfig.ps1` script to authenticate. Maybe we can use the same setup for the unit tests?

stijnmoreels:
> @stijnmoreels any ideas on how to go about this? I see that the integration tests contain a `Connect-AzAccountFromConfig.ps1` script to authenticate. Maybe we can use the same setup for the unit tests?

Well, we shouldn't call any 'authenticate' methods in our unit tests. We wouldn't want them to fail bc there's no external connection with APIM, for example.

I think we should approach this the same way as we did with the Azure DevOps where we use the built-in access token instead of authenticating ourselves. Maybe here, we could pass in the access token as one of the parameters? We already have like a function called `Get-AzCachedAccessToken` which does exactly the same.

```powershell
$accessToken = Get-AzCachedAccessToken
Remove-AzApiManagementSoftDeletedService -Name "my-apim" -AccessToken $accessToken
```

That way, we could stub-out the `Invoke-RestMethod` with our own HTTP response. Would that work?

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
...
+    [Parameter(Mandatory = $false)][string] $AccessToken = ""
+)
+
+if($SubscriptionId -eq "" -or $AccessToken -eq ""){
+    # Request accessToken in case the script contains records
+    $token = Get-AzCachedAccessToken
+
+    $AccessToken = $token.AccessToken
+    $SubscriptionId = $token.SubscriptionId
+}
```
stijnmoreels, line 13: I was wondering why we did this, or if this comment is actually correct. Seems to me that it should be `...in case the script contains no records.` 😅 But I saw the same thing indeed with Logic Apps. Hmm. 🤔

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
...
+
+$authHeader = @{
+   'Authorization'='Bearer ' + $AccessToken
+}
+
+Write-Host "Checking if the API Management instance with name '$Name' is listed as a soft deleted service"
+$getUri = 'https://management.azure.com/subscriptions/{0}/providers/Microsoft.ApiManagement/deletedservices?api-version=2021-08-01' -f $SubscriptionId
```
stijnmoreels, line 20: We can make this API version also an optional parameter. Like we did with the Logic Apps implementation, too.

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
...
+$deletedServices = (Invoke-RestMethod -Method GET -Uri $getUri -Headers $authHeader)
+
+if ($deletedServices.value.count -eq 0 -or ($deletedServices.value | Where-Object name -eq $Name).count -eq 0) {
+    throw "API Management instance with name '$Name' is not listed as a soft deleted service and therefore it cannot be removed"
```
stijnmoreels, line 24: Great logging in this script! 🥇 Will help a lot.

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
stijnmoreels, line 20: We should probably also make this URL configurable based on the cloud. (Like, global cloud, Germany...). I think we already have something like this in Logic Apps, so you can check. Can be done in a separate PR, though.

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
...
+
+    $AccessToken = $token.AccessToken
+    $SubscriptionId = $token.SubscriptionId
+}
+
+$authHeader = @{
+   'Authorization'='Bearer ' + $AccessToken
+}
+
+Write-Host "Checking if the API Management instance with name '$Name' is listed as a soft deleted service"
```
stijnmoreels, line 19: This is more a 'verbose' call, no?

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
...
+$getUri = 'https://management.azure.com/subscriptions/{0}/providers/Microsoft.ApiManagement/deletedservices?api-version=2021-08-01' -f $SubscriptionId
+$deletedServices = (Invoke-RestMethod -Method GET -Uri $getUri -Headers $authHeader)
+
+if ($deletedServices.value.count -eq 0 -or ($deletedServices.value | Where-Object name -eq $Name).count -eq 0) {
+    throw "API Management instance with name '$Name' is not listed as a soft deleted service and therefore it cannot be removed"
+}
+
+Write-Host "API Management instance has been found for name '$Name' as a soft deleted service"
+
+Write-Host "Removing the soft deleted API Management instance '$Name'"
```
stijnmoreels, line 29: This is more a verbose call, no?

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,38 @@
+Param(
+    [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
+    [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
+    [Parameter(Mandatory = $false)][string] $AccessToken = ""
+)
+
+if($SubscriptionId -eq "" -or $AccessToken -eq ""){
+    # Request accessToken in case the script contains records
+    $token = Get-AzCachedAccessToken
```
stijnmoreels, line 9: Great stuff in re-using existing functionality. implementation here looks real nice and clean! 💯

src/Arcus.Scripting.Management/Scripts/Restore-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,46 @@
...
+    $SubscriptionId = $token.SubscriptionId
+}
+
+$authHeader = @{
+   'Authorization'='Bearer ' + $AccessToken
+}
+
+Write-Host "Checking if the API Management instance with name '$Name' is listed as a soft deleted service"
+$getUri = 'https://management.azure.com/subscriptions/{0}/providers/Microsoft.ApiManagement/deletedservices?api-version=2021-08-01' -f $SubscriptionId
+$deletedServices = (Invoke-RestMethod -Method GET -Uri $getUri -Headers $authHeader)
```
stijnmoreels, line 21: Is it an idea to move this to a separate script as it's the same for both scripts? Or would that make things too complicated? Again, not critical for me, though, can be seen as an improvement later.

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
```diff
@@ -0,0 +1,199 @@
...
+
+                # Act
+                { 
+                   Remove-AzApiManagementSoftDeletedService -Name $Name
+                } | Should -Throw -ExpectedMessage "API Management instance with name '$Name' is not listed as a soft deleted service and therefore it cannot be removed"
+
+                # Assert
+                Assert-VerifiableMock
+                Assert-MockCalled Get-AzCachedAccessToken -Scope It -Times 1
+            }
```
stijnmoreels, line 30: If it's 'not existing' APIM resources we're testing, this can be an integration test, too right?

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
```diff
@@ -0,0 +1,199 @@
...
+                    if ($Method -eq "Get") {
+                       return [pscustomobject] @{
+                        value = @([ordered] @{
+                                id = "subscriptions/########-####-####-####-############/providers/Microsoft.ApiManagement/locations/westeurope/deletedservices/$Name";
+                                name = $Name;
+                                location = "West Europe";
+                            })
+                        };
+                    } else {
+                        Throw 'some error'
```
stijnmoreels, line 52:
```suggestion
                        throw 'some error'
```

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
```diff
@@ -0,0 +1,199 @@
...
```
stijnmoreels, line 148:
```suggestion
                        throw 'some error'
```

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
```diff
@@ -0,0 +1,199 @@
...
+                    if ($Method -eq "Get") {
+                       return [pscustomobject] @{
+                        value = @([ordered] @{
+                                id = "subscriptions/########-####-####-####-############/providers/Microsoft.ApiManagement/locations/westeurope/deletedservices/$Name";
+                                name = $Name;
+                                location = "West Europe";
+                            })
+                        };
+                    } else {
+                        return $null
```
stijnmoreels, line 183: Is it an idea to also assert on the HTTP method here? That it should always be DELETE? And in the restore-case PUT?

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,37 @@
...
+ .Description
+  Restore a soft deleted API Management instance.
+
+ .Parameter Name
+  The name of the Azure API Management instance which has been soft deleted and will be restored.
+#>
+function Restore-AzApiManagementSoftDeletedService {
+    param(
+        [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required")
+    )
```
stijnmoreels, line 34:
```suggestion
 .Parameter Name
  The name of the Azure API Management instance which has been soft deleted and will be restored.
 
 .Parameter SubscriptionId
  [Optional] The Id of the subscription containing the Azure API Management instance. When not provided, it will be retrieved from the current context (Get-AzContext).
  
 .Parameter AccessToken
  [Optional] The access token to be used to restore the Azure API Management instance.   
#>
function Restore-AzApiManagementSoftDeletedService {
    param(
        [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
        [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
        [Parameter(Mandatory = $false)][string] $AccessToken = ""
    )
```

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,37 @@
...
+
+    . $PSScriptRoot\Scripts\Restore-AzApiManagementSoftDeletedService.ps1 -Name $Name
```
stijnmoreels, line 36:
```suggestion
    . $PSScriptRoot\Scripts\Restore-AzApiManagementSoftDeletedService.ps1 -Name $Name -SubscriptionId $SubscriptionId -AccessToken $AccessToken
```

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,37 @@
...
+
+ .Parameter Name
+  The name of the Azure API Management instance which has been soft deleted and will be permanently removed.
+#>
+function Remove-AzApiManagementSoftDeletedService {
+    param(
+        [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required")
+    )
+
+    . $PSScriptRoot\Scripts\Remove-AzApiManagementSoftDeletedService.ps1 -Name $Name
```
stijnmoreels, line 16:
```suggestion
 .Parameter Name
  The name of the Azure API Management instance which has been soft deleted and will be permanently removed.
  
 .Parameter SubscriptionId
  [Optional] The Id of the subscription containing the Azure API Management instance. When not provided, it will be retrieved from the current context (Get-AzContext).
  
 .Parameter AccessToken
  [Optional] The access token to be used to restore the Azure API Management instance.   
#>
function Remove-AzApiManagementSoftDeletedService {
    param(
        [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
        [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
        [Parameter(Mandatory = $false)][string] $AccessToken = ""
    )

    . $PSScriptRoot\Scripts\Remove-AzApiManagementSoftDeletedService.ps1 -Name $Name -SubscriptionId $SubscriptionId -AccessToken $AccessToken
```

docs/preview/02-Features/powershell/azure-management.md
```diff
@@ -0,0 +1,65 @@
+
+---
+title: "Azure Management"
+layout: default
+---
+
+# Azure Management
```
stijnmoreels, line 7: Clean and to-the-point as always, 👍 Great job!!

docs/preview/02-Features/powershell/azure-management.md
```diff
@@ -0,0 +1,65 @@
...
+To have access to the following features, you have to import the module:
+
+```powershell
+PS> Install-Module -Name Arcus.Scripting.Management
+```
+
+## Removing a soft deleted API Management instance
+
+Removes a soft deleted API Management instance. 
+For more information on API Management and soft deletion see [here](https://docs.microsoft.com/en-us/azure/api-management/soft-delete#soft-delete-behavior).
```
stijnmoreels, line 26: Probably good that we have some references, indeed. Super!

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,37 @@
...
+}
+
+Export-ModuleMember -Function Set-AzIntegrationAccountSchemas
```
stijnmoreels, line 19:
```suggestion
Export-ModuleMember -Function Remove-AzApiManagementSoftDeletedService
```

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,37 @@
...
+ .Parameter Name
+  The name of the Azure API Management instance which has been soft deleted and will be restored.
+#>
+function Restore-AzApiManagementSoftDeletedService {
+    param(
+        [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required")
+    )
+
+    . $PSScriptRoot\Scripts\Restore-AzApiManagementSoftDeletedService.ps1 -Name $Name
+}
```
stijnmoreels, line 37:
```suggestion
}

Export-ModuleMember -Function Restore-AzApiManagementSoftDeletedService
```

stijnmoreels (changes requested):
This is already shaping to be a great new management package. Excited to have this included in our new version. I have some suggestions and comments, but not all are blocking for this PR.

These are non-blockers for me:
- [ ] Make API version configurable (see Logic Apps)
- [ ] Make management URL's configurable with Azure cloud (see Logic Apps)
- [ ] Extract duplicate call to retrieve deleted APIM resources to separate script (see Logic Apps management URL script)

Thx for this already!

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: You are right, should be `contains no records`, you want me to change this and in the Logic Apps part as well?

stijnmoreels: Maybe that will clutter things in this PR? You can choose 😄 . I won't mind if we do this in another PR.

pim-simons: As it is just a comment, lets do it in a seperate PR later on 👍🏻

pim-simons: I will change it for this code and fix the Logic Apps later on in a seperate PR 👍🏻

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: Will set it up as an optional parameter and set it as default on `2021-08-01`.
I do wonder how we will handle newer api-versions that might have breaking changes, different to Logic Apps we are using a body in one of the http calls so this might be a bit more subject to breaking changes. @stijnmoreels what is your view on this?

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: Thanks for your suggestion on this 👍🏻

docs/preview/02-Features/powershell/azure-management.md
pim-simons: Thanks!

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
stijnmoreels: Ah, you have a point there. Hmm, 🤔 . Since it's already a 'fairly' recent API version, we may choose to not have an optional parameter, then. And only make the Azure cloud configurable. Hmm, that's somewhat the trade-off I guess here, bc there's no built-in PS commands for us to do this.
Yeah, if we don't expose the API version, then we may be able to have more control to update the body/URL when breaking changes occur?
Adding @fgheysels @mbraekman to this.

pim-simons: We can make it optional and put in some documentation that states we have tested support up till api-version `2021-08-01`.
Adding this optional parameter does have the benefit of not having to continually update the script whenever new api-versions are available.

stijnmoreels: Aha, yeah, that seems like a good compromise between the two things. Let's do that! 👍

stijnmoreels: It will be better that way bc it will be more streamlined with our other scripts.

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: Should be fixed now, see latest commit

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: Added as an optional parameter 👍🏻 see latest commit

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
pim-simons: Agreed, changed it in the latest commit

src/Arcus.Scripting.Management/Scripts/Restore-AzApiManagementSoftDeletedService.ps1
pim-simons: I moved the getting of the soft deleted instances to a seperate script, see latest commit

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
pim-simons: Good one 👍🏻 added as an integration test

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.Management.tests.ps1
pim-simons: Added this to the assertion on the mock

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -18,10 +18,12 @@ function Remove-AzApiManagementSoftDeletedService {
     param(
         [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
         [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
-        [Parameter(Mandatory = $false)][string] $AccessToken = ""
+        [Parameter(Mandatory = $false)][string] $EnvironmentName = "AzureCloud",
+        [Parameter(Mandatory = $false)][string] $AccessToken = "",
+        [Parameter(Mandatory = $false)][string] $ApiVersion = "2021-08-01"
```
stijnmoreels, line 23: Ok, now I'm nitpicking 😅 . The access token and subscription ID belongs more together? Maybe the access token can be moved a place above the environment name?

src/Arcus.Scripting.Management/Scripts/Get-AzApiManagementResourceManagementUrl.ps1
```diff
@@ -0,0 +1,34 @@
...
+        $supportedEnvironments = ""
+
+        foreach($env in $environments){
+            if($supportedEnvironments.Length -eq 0) {
+                $supportedEnvironments += $env
+            }
+            else{
+                $supportedEnvironments += ", " + $env
+            }
+        }
```
stijnmoreels, line 20:
```suggestion
try {
    $resourceManagerUrl = ""   
    
    $environments = (Get-AzEnvironment).Name
    if ($EnvironmentName -notin $environments) {
        $supportedEnvironments = ""

        foreach ($env in $environments) {
            if ($supportedEnvironments.Length -eq 0) {
                $supportedEnvironments += $env
            }
            else {
                $supportedEnvironments += ", " + $env
            }
        }
```

src/Arcus.Scripting.Management/Scripts/Get-AzApiManagementResourceManagementUrl.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+        Write-Error "Unrecognized environment specified. Supported values are: $supportedEnvironments"
+    }
+
+    $resourceManagerUrl = (Get-AzEnvironment -Name $EnvironmentName).ResourceManagerUrl
+
+    return $resourceManagerUrl
+}
+catch {
+    Write-Warning "Failed to retrieve the resource management endpoint."
```
stijnmoreels, line 30:
```suggestion
    Write-Warning "Failed to retrieve the resource management endpoint"
```
I think our messages don't have dots?

src/Arcus.Scripting.Management/Scripts/Get-AzApiManagementSoftDeletedResources.ps1
```diff
@@ -0,0 +1,20 @@
+Param(
```
stijnmoreels, line 1:
```suggestion
param(
```

src/Arcus.Scripting.Management/Scripts/Restore-AzApiManagementSoftDeletedService.ps1
```diff
@@ -1,11 +1,13 @@
 Param(
```
stijnmoreels, line 1:
```suggestion
param(
```

src/Arcus.Scripting.Management/Scripts/Restore-AzApiManagementSoftDeletedService.ps1
```diff
@@ -1,11 +1,13 @@
...
     [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
     [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
-    [Parameter(Mandatory = $false)][string] $AccessToken = ""
+    [Parameter(Mandatory = $false)][string] $EnvironmentName = "AzureCloud",
+    [Parameter(Mandatory = $false)][string] $AccessToken = "",    
+    [Parameter(Mandatory = $false)][string] $ApiVersion = "2021-08-01"
 )
 
 if($SubscriptionId -eq "" -or $AccessToken -eq ""){
```
stijnmoreels, line 9:
```suggestion
if ($SubscriptionId -eq "" -or $AccessToken -eq "") {
```

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,34 @@
+Param(
```
stijnmoreels, line 1:
```suggestion
param(
```

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,34 @@
...
+    [Parameter(Mandatory = $true)][string] $Name = $(throw "Name of the API Management instance is required"),
+    [Parameter(Mandatory = $false)][string] $SubscriptionId = "",
+    [Parameter(Mandatory = $false)][string] $EnvironmentName = "AzureCloud",
+    [Parameter(Mandatory = $false)][string] $AccessToken = "",
+    [Parameter(Mandatory = $false)][string] $ApiVersion = "2021-08-01"
+)
+
+if($SubscriptionId -eq "" -or $AccessToken -eq ""){
```
stijnmoreels, line 9:
```suggestion
if ($SubscriptionId -eq "" -or $AccessToken -eq "") {
```

src/Arcus.Scripting.Management/Scripts/Remove-AzApiManagementSoftDeletedService.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+    $AccessToken = $token.AccessToken
+    $SubscriptionId = $token.SubscriptionId
+}
+
+$authHeader = @{
+   'Authorization'='Bearer ' + $AccessToken
+}
+
+$resourceManagerUrl = . $PSScriptRoot\Get-AzApiManagementResourceManagementUrl.ps1 -EnvironmentName $EnvironmentName
```
stijnmoreels, line 21: Wauw, look at that. As clean as it gets 😉 .

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
```diff
@@ -0,0 +1,59 @@
...
+
+ .Parameter Name
+  The name of the Azure API Management instance which has been soft deleted and will be restored.
+ 
+ .Parameter SubscriptionId
+  [Optional] The Id of the subscription containing the Azure API Management instance. When not provided, it will be retrieved from the current context (Get-AzContext).
+  
+ .Parameter AccessToken
+  [Optional] The access token to be used to restore the Azure API Management instance.   
+#>
```
stijnmoreels, line 46: The environment name and API versions will probably have to be added here. You can probably copy-paste it more or less from the Logic Apps variant, I think.

docs/preview/02-Features/powershell/azure-management.md
```diff
@@ -0,0 +1,73 @@
...
+| `Name`           | yes       | The name of the API Management instance that has been soft deleted.                                        |
+| `SubscriptionId` | no        | The Id of the subscription containing the Azure API Management instance.                                   |
+|                  |           | When not provided, it will be retrieved from the current context (Get-AzContext).                          |
+| `EnvironmentName`| no        | The name of the Azure environment where the Azure API Management instance resides. (default: `AzureCloud`) |
+| `AccessToken`    | no        | The access token to be used to restore the Azure API Management instance.                                  |
+|                  |           | When not provided, it will be retrieved from the current context (Get-AzContext).                          |
+| `ApiVersion `    | no        | The version of the management API to be used.  (default: `2021-08-01`)                                     |
+
+> :bulb: The `ApiVersion` has successfully been tested with version `2021-08-01`.
```
stijnmoreels, line 64: Super!

stijnmoreels (approved):
I'll already approve bc it's already almost done 😄 .
I have some last remarks on super super small things 🤏 .

Thx for this! 🥇

src/Arcus.Scripting.Management/Arcus.Scripting.Management.psm1
stijnmoreels: Ah, nevermind. I see that we do this the same way as with Logic Apps. My bad! It's god as it is.

stijnmoreels:
Woops, sorry, I saw that we did this also in Logic Apps. You can revert the order back 😅 . I was mislead by the subscription ID. The subscription ID and environment name are indeed more linked together. Sorry about that.

pim-simons:
> Woops, sorry, I saw that we did this also in Logic Apps. You can revert the order back 😅 . I was mislead by the subscription ID. The subscription ID and environment name are indeed more linked together. Sorry about that.

No worries, will change it back 😉

stijnmoreels (approved):

pim-simons:
> 

Thanks for the great reviews @stijnmoreels! Greatly appreciated 👍🏻

PR #281, opened 2022-03-16
test: removal of purging keyvault key

pim-simons:
Since we are now using unique keynames we don't need to purge the keyvault keys anymore for the private keys in the private key tests. This was causing tests to sometimes fail as the purging happened to fast after initial deletion.


stijnmoreels (approved):
Aha aha, great! Wasn't I think fully aware we did this. Nice catch! 👍

Issue #284, opened 2022-03-18
Doc: improve performance of using Arcus libs in DevOps pipelines


fgheysels:
We've heard multiple comments from people that are using Arcus.Scripting in their DevOps pipelines that they're frustrated by the time it takes to install an Arcus package in their pipeline.
I personally have experienced this myself:  In one of my steps in the pipeline, I want to make use of functionality provided by Arcus, and I execute for instance this command:

```powershell
Install-Module -Name Arcus.Scripting.KeyVault -Force
```
This step takes over 5 minutes to execute, which is way too long, and people are being frustrated about that.
I think it takes 5 minutes, because the `-Force` parameter makes sure that all dependent modules are also being installed.

I suggest that we add some tips regarding this in our documentation.
Instead of using the `-Force` parameter, we could suggest to use the commands below.  It will drastically increase performance 👍🏻 

```powershell
Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
Install-Module -Name Arcus.Scripting.KeyVault -AllowClobber
```

stijnmoreels:
Aha aha! Yeah, if this combination indeed helps faster installation. Why not? We don't mention the `-Force` switch, though, in our docs.
This is rather something typical PowerShell, and not per-se Arcus. I image the same would be true for all other PS modules out there. But, yeah, maybe people are only copy-pasting what they see 😅 . This is maybe even something for the academy, on how to work with PowerShell in DevOps.

stijnmoreels:
@spasal , great that you already worked on this! Mind that assignment of issues/work is also important, as someone else could be doing the same thing without anyone dedicated assigned on the issue. If the issue is of your own making and you want to propose something, by all means! But an other issue from someone else, can maybe lead to duplicate work, without knowing of each other. Just a quick ping, and we can assign you 😉 . We wouldn't want to disappoint anyone that their work and time was already done by someone else. Super super much thanks for your effort on Arcus! Very very much appreciated! ❤️

fgheysels:
I think we can close this issue ?

fgheysels:
I'd like to reopen this.
We have already updated our documentation for this, but only the specific documentation on how to install the AzureDevops module contains this information.
You can use other modules in DevOps pipelines as well, so I'd like to have this information more on a 'global' section in the [docs](https://scripting.arcus-azure.net/). 

Why not add a section between 'Welcome' and 'Features' that is called 'Guidelines' , where we put some best practices / advise ?

[...]

PR #285, opened 2022-03-18
docs: updated Azure DevOps feature docs regarding installing Arcus packages in pipelines

spasal:
Updated the Azure DevOps docs on how to install the Arcus packages inside your pipeline.
Previously, using -Force, this step could take up to 5 minutes.
Now, by registering the repository and using -AllowClobber step takes one minute.

Relates to #284

### DIFF
--- a/docs/preview/02-Features/powershell/azure-devops.md
+++ b/docs/preview/02-Features/powershell/azure-devops.md
@@ -88,7 +88,7 @@ This function is intended to be used from an Azure DevOps pipeline. Internally, 
 One of the environment variables that is used, is the `SYSTEM_ACCESSTOKEN` variable. However, due to safety reasons this variable is not available out-of-the box.
 To be able to use this variable, it must be explicitly added to the environment-variables.
 
-> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+> ⚠ When you are using a Linux agent (which is recommended), you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
 
 Example of how to use this function in an Azure DevOps pipeline:
 
@@ -143,7 +143,7 @@ PS> Set-AzDevOpsArmOutputsToPipelineVariables -ArmOutputsEnvironmentVariableName
 **Azure DevOps Example**
 This function is intended to be used from an Azure DevOps pipeline.
 
-> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+> ⚠ When you are using a Linux agent (which is recommended), you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
 
 Example of how to use this function in an Azure DevOps pipeline:
 
--- a/docs/preview/02-Features/powershell/azure-devops.md
+++ b/docs/preview/02-Features/powershell/azure-devops.md
@@ -88,7 +88,9 @@ This function is intended to be used from an Azure DevOps pipeline. Internally, 
 One of the environment variables that is used, is the `SYSTEM_ACCESSTOKEN` variable. However, due to safety reasons this variable is not available out-of-the box.
 To be able to use this variable, it must be explicitly added to the environment-variables.
 
-> ⚠ When you are using a Linux agent (which is recommended), you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+
+> We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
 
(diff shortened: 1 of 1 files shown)


fgheysels (approved):
Thanks for changing this so quickly in these code-samples!

However, I think we should immediately close the related issue.   In my humble opinion, we should have some docs which explicitly state that you can install the package this way

spasal:
> Thanks for changing this so quickly in these code-samples!
> 
> However, I think we should immediately close the related issue. In my humble opinion, we should have some docs which explicitly state that you can install the package this way

I have added a recommendation to use a Linux agent instead of a Windows agent in the text. Installing the package takes multiple minutes with Windows, but less than one minute for Linux.

docs/preview/02-Features/powershell/azure-devops.md
```diff
@@ -88,7 +88,7 @@ This function is intended to be used from an Azure DevOps pipeline. Internally,
 One of the environment variables that is used, is the `SYSTEM_ACCESSTOKEN` variable. However, due to safety reasons this variable is not available out-of-the box.
 To be able to use this variable, it must be explicitly added to the environment-variables.
 
-> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+> ⚠ When you are using a Linux agent (which is recommended), you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
```
fgheysels, line 91: Using a Linux agent is not always possible.   For instance when you want to execute a DacPac task, or if you want to build a Windows container.
I'd change this to something like this:

```suggestion
> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.

> We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
```

docs/preview/02-Features/powershell/azure-devops.md
```diff
@@ -142,7 +143,7 @@ PS> Set-AzDevOpsArmOutputsToPipelineVariables -ArmOutputsEnvironmentVariableName
 **Azure DevOps Example**
 This function is intended to be used from an Azure DevOps pipeline.
 
-> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
+> ⚠ When you are using a Linux agent (which is recommended), you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
```
spasal, line 146:
```suggestion
> ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.

> We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
```

stijnmoreels:
I think @fgheysels meant, "we should _not_ immediately close the issue" 😅 But Alex already took care of this by using 'Related to' instead. So, all is good. 👍

docs/preview/02-Features/powershell/azure-devops.md
```diff
@@ -90,6 +90,8 @@ To be able to use this variable, it must be explicitly added to the environment-
 
 > ⚠ When you are using a Linux agent, you need to pass other environment variables that you want to use as well, because these are not available. To be able to use the `ArmOutputs` environment variable, it must be explicitly added to the environment-variables.
 
+> We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
```
stijnmoreels, line 93:
```suggestion
> 💡 We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
```

docs/preview/02-Features/powershell/azure-devops.md
```diff
@@ -144,6 +147,8 @@ This function is intended to be used from an Azure DevOps pipeline.
...
```
stijnmoreels, line 150:
```suggestion
> 💡 We have seen a much better performance when using Linux agents, and would recommend using Linux agents when possible.
```

stijnmoreels (approved):
Great stuff! 👍 Thx for picking this up so quickly! Shows great commitment to Arcus. Amazing to receive these updates from you. 😉

fgheysels:
> I think @fgheysels meant, "we should _not_ immediately close the issue" 😅 But Alex already took care of this by using 'Related to' instead. So, all is good. 👍

I actually meant that we might need to add this somewhere in a more 'general' section of the docs as well ?

pim-simons (approved):

pim-simons (changes requested):
@spasal looks very good! As @fgheysels mentioned https://github.com/arcus-azure/arcus.scripting/pull/285#issuecomment-1075299360 it would be good to add the same code-sample to lines 21 to 23. 
Could you change it on lines lines 21 to 23 there as well to: 
```
Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
Install-Module -Name Arcus.Scripting.DevOps -Repository PSGallery -AllowClobber
```

spasal:
> @spasal looks very good! As @fgheysels mentioned [#285 (comment)](https://github.com/arcus-azure/arcus.scripting/pull/285#issuecomment-1075299360) it would be good to add the same code-sample to lines 21 to 23. Could you change it on lines lines 21 to 23 there as well to:
> 
> ```
> Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
> Install-Module -Name Arcus.Scripting.DevOps -Repository PSGallery -AllowClobber
> ```

Ahah, missed that one. Updated it, thanks for the suggestion!

PR #298, opened 2022-04-22
docs: use 'app service' io 'integration account' fix

stijnmoreels:
Should be 'App Service' probably, instead of 'Integration Account. Should be fixed now.


pim-simons (approved):
Good catch 👍🏻 probably a copy-paste error 🙈

PR #305, opened 2022-05-06
chore: add readme to ps pkg (🚧)

stijnmoreels:
Adds the README file to the build packages.

Relates to https://github.com/arcus-azure/arcus/issues/209


stijnmoreels:
Added the README.md file to the package. As seen on MyGet, this file is now part of the package, but is not shown in MyGet. If I look at the other packages on the PowerShell Gallery, there also doesn't seem to be any README, so I'll just close this one.

PR #310, opened 2022-05-17
chore: improve local pester test runs

### LINKED ISSUE #309: PowerShell scripts to easily run the Pester tests
**Is your feature related to a problem? Please subscribe**
When running the Pester tests locally, we come across some problems due to the fact that the module version is not valid (we use a token that we replace in the YAML build) and that the application settings are not filled out (we use the general ones and should use local application settings if we run the tests locally).

**What solution do you propose?**
We should make sure that we:
* Make sure that we provide a valid module version that we still can replace during the YAML build
* Make sure that we take any local application settings into account when running the Pester tests locally

### Discussed in https://github.com/arcus-azure/arcus.scripting/discussions/307

<div type='discussions-op-text'>

<sup>Originally posted by **stijnmoreels** May 13, 2022</sup>
Currently, when you want to run the Pester unit/integration tests. You need to do some manually tasks first locally, like:
* Setting a valid root module version of related modules
* Setting the appsettings to the local variant in case of integration tests

Wouldn't it be nice if we could have some PS scripts in our repo that does this for us locally? So, if we want for example run unit tests, we could just:

```powershell
PS> .\Run-UnitTests.ps1
PS> .\Run-UnitTests.ps1 -Module Arcus.Scripting.LogicApps
PS> .\Run-IntegrationTests.ps1 -Module Arcus.Scripting.ARM
``` 

Just thinking out-loud. We could place our scripts in the repo so everyone has access to them and can run them. The scripts would need to set a valid version to the modules and replace the `appsettings.json` to `appsettings.local.json` before the `Invoke-Pester` command could be run. Also, these changes should be reverted once the test run is done.</div>

stijnmoreels:
Improve the Pester test runs by using a valid module version instead of tokens and taking into account any local appsettings.

Closes #309


build/templates/replace-tokens.yml
```diff
@@ -0,0 +1,29 @@
+steps:
+  - powershell: |
+      Get-ChildItem -Path ./src -Filter *.psd1 -Recurse |
+        % { Write-Host "Replace '0.0.0' to 'Package.Version' in $($_.FullName)"
+            $contents = Get-Content $_.FullName 
+            $contents = $contents -replace "ModuleVersion = '0.0.0'", "ModuleVersion = '#{Package.Version}#'"
```
pim-simons, line 6: Should we only replace when it is `0.0.0`? Or just replace whatever the version is?

pim-simons (approved):
Awesome work, this will make our developing life quite a bit easier 👍🏻

build/templates/replace-tokens.yml
stijnmoreels: Ah, you make a good point. Was first hesitate to do a regular expression check (like `\d\.\d\.\d`) bc it would also replace the versions in the `RequiredModules` section, but since we take in the `ModuleVersion` we could indeed do that. Nice catch! 👍

pim-simons (approved):

PR #312, opened 2022-06-20
test: simplify arm integration tests

### LINKED ISSUE #279: Investigate in loading ARM test files instead of keeping original files in source control
Currently, we've chosen for a clear separation of ARM test files for input/output so there's a clear before/after on disk. But, maybe with a clever value loading, we can generate the input file and reduce the amount of test files.

Investigate in whether we can load the value to a test file, instead of using the file itself. Therefore we only have to delete the generated test file instead of replacing the contents of the input file.

**Additional context**
Related to Pieter V's PR's question.

stijnmoreels:
Simplify ARM integration tests by removing the need for an `-org.json` original JSON file but re-using the initial content of the ARM template file as 'original contents' which can be used during the tear-down of each test, keeping a clean environment after each test.

Closes #279


pim-simons:
@stijnmoreels  this test seems to be failing:
```
[-] Arcus ARM integration tests.ARM injection.Replaces absolute file path with file contents as JSON object (linux) 51ms (49ms|2ms)
 CommandNotFoundException: The term 'OutFile' is not recognized as a name of a cmdlet, function, script file, or executable program.
 Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
 at <ScriptBlock>, /home/vsts/work/1/s/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.ARM.tests.ps1:83
```

stijnmoreels:
> 

Aha! Was using `OutFile` and should've been `Out-File`. 😄 Couldn't test this locally as I'm running Windows.
Thx for the snippet! 👍

pim-simons (approved):
Looks good to me, thanks 🥇

Issue #220, opened 2021-09-14
Add extra validation when creating a new Azure API Management context


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
In the Azure API Management-specific scripts, we mostly create a new context with the APIM details without any validation. This can cause runtime failures afterwards in the following lines.

**Describe the solution you'd like**
Add an additional line of validation to verify if the provided APIM details point towards a valid/existing instance.

**Describe alternatives you've considered**

**Additional context**
Mentioned in #206

stijnmoreels:
We'll need some other remote call to verify the resource group and service name separately, because the context returned from `New-AzApiManagementContext`  doesn't contain any validation. Only a data object: [`PsApiManagementContext`](https://github.com/Azure/azure-powershell/blob/main/src/ApiManagement/ApiManagement.ServiceManagement/Models/PsApiManagementContext.cs).

PR #206, opened 2021-07-30
feat: upload public ca certificate to az api management

stijnmoreels:
Adds new PS script to Azure API Management module to upload a public CA certificate to the Azure API Management certificate store.

Closes #134

PR #317, opened 2022-08-16
feat: Add APIM context validation

pim-simons:
Added context validation for Azure API Management. 

Closes https://github.com/arcus-azure/arcus.scripting/issues/220

### DIFF
--- a/src/Arcus.Scripting.ApiManagement/Scripts/Create-AzApiManagementApiOperation.ps1
+++ b/src/Arcus.Scripting.ApiManagement/Scripts/Create-AzApiManagementApiOperation.ps1
@@ -10,10 +10,11 @@ param(
    [Parameter(Mandatory = $false)][string] $PolicyFilePath = ""
 )
 
-$apimContext = New-AzApiManagementContext -ResourceGroupName $ResourceGroupName -ServiceName $ServiceName
-if ($apimContext -eq $null) {
-    throw "Unable to find the Azure API Management Instance '$ServiceName' in resource group $ResourceGroupName"
+$apim = Get-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName
+if ($apim -eq $null) {
+    throw "Unable to find the Azure API Management Instance $ServiceName in resource group $ResourceGroupName"
 }
+$apimContext = New-AzApiManagementContext -ResourceGroupName $ResourceGroupName -ServiceName $ServiceName
 
 New-AzApiManagementOperation -Context $apimContext -ApiId $ApiId -OperationId $OperationId -Name $OperationName -Method $Method -UrlTemplate $UrlTemplate -Description $Description
 Write-Host "New API operation '$OperationName' on API Management instance was added."
--- a/src/Arcus.Scripting.ApiManagement/Scripts/Create-AzApiManagementApiOperation.ps1
+++ b/src/Arcus.Scripting.ApiManagement/Scripts/Create-AzApiManagementApiOperation.ps1
@@ -11,6 +11,9 @@ param(
 )
 
 $apimContext = New-AzApiManagementContext -ResourceGroupName $ResourceGroupName -ServiceName $ServiceName
+if ($apimContext -eq $null) {
+    throw "Unable to find the Azure API Management Instance '$ServiceName' in resource group $ResourceGroupName"
+}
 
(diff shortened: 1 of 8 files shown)


stijnmoreels:
Ah, was I wrong in assuming that this comment: https://github.com/arcus-azure/arcus.scripting/issues/220#issuecomment-918892379 was outdated or something? I remember that when looking at the code, it only created a custom PS object, which of course is never `$null` , I figured that we needed some remote call to verify if the context represents a valid APIM.

pim-simons:
> Ah, was I wrong in assuming that this comment: [#220 (comment)](https://github.com/arcus-azure/arcus.scripting/issues/220#issuecomment-918892379) was outdated? I remember that when looking at the code, it only created a custom PS object, which of course is never `$null` , I figured that we needed some remote call to verify if the context represents a valid APIM.

I mostly just took this from this comment here: https://github.com/arcus-azure/arcus.scripting/pull/206/files#diff-0479d3754dad91471518225257039759e3809046238bd16748e1e71adadc7eac: 
> This is indeed something that perhaps should be implemented on other scripts as well, just to have some additional line of validation to confirm whether the provided APIM-details are actually pointing towards a valid/existing instance.

So basically I added this part to the code where the APIM context is retrieved:
``` powershell
if ($apimContext -eq $null) {
    throw "Unable to find the Azure API Management Instance '$ServiceName' in resource group $ResourceGroupName"
}
```

But if I understand correctly we need some kind of remote call to validate? I wasn't onboard yet when the original discussion took place, so maybe I missed something 😉

stijnmoreels:
> But if I understand correctly we need some kind of remote call to validate? I wasn't onboard yet when the original discussion took place, so maybe I missed something 😉

Yeah, seems like it. You can see it here: https://github.com/Azure/azure-powershell/blob/4668bff1cc4ed6510d9b0d4c6ec289ab27f04241/src/ApiManagement/ApiManagement.ServiceManagement/Commands/NewAzureApiManagementContext.cs

Mostly, the context is only created. Maybe there exists a `Contains API context` command somewhere to verify this, but to-be-analyzed. Maybe there is some kind of check hidden somewhere in the API Management commands.

pim-simons:
Seems like using the approach used [here](https://github.com/arcus-azure/arcus.scripting/blob/master/src/Arcus.Scripting.ApiManagement/Scripts/Upload-AzApiManagementSystemCertificate.ps1#L14) might be a way to go.
This uses [Get-AzApiManagement](https://docs.microsoft.com/en-us/powershell/module/az.apimanagement/get-azapimanagement?view=azps-8.2.0) which gets the specific API Management instance so we can check if it actually exists.
@stijnmoreels what do you think?

pim-simons:
> > Seems like using the approach used [here](https://github.com/arcus-azure/arcus.scripting/blob/master/src/Arcus.Scripting.ApiManagement/Scripts/Upload-AzApiManagementSystemCertificate.ps1?rgh-link-date=2022-08-17T09%3A00%3A55Z#L14) might be a way to go. This uses [Get-AzApiManagement](https://docs.microsoft.com/en-us/powershell/module/az.apimanagement/get-azapimanagement?view=azps-8.2.0) which gets the specific API Management instance so we can check if it actually exists. @stijnmoreels what do you think?
> 
> Aha, yeah, that could be a good alternative 👍

Appplied in [`1b62d75` (#317)](https://github.com/arcus-azure/arcus.scripting/pull/317/commits/1b62d758db6566bf0a948d42b022dbcaccb49a35)

stijnmoreels:
> > > Seems like using the approach used [here](https://github.com/arcus-azure/arcus.scripting/blob/master/src/Arcus.Scripting.ApiManagement/Scripts/Upload-AzApiManagementSystemCertificate.ps1?rgh-link-date=2022-08-17T09%3A00%3A55Z#L14) might be a way to go. This uses [Get-AzApiManagement](https://docs.microsoft.com/en-us/powershell/module/az.apimanagement/get-azapimanagement?view=azps-8.2.0) which gets the specific API Management instance so we can check if it actually exists. @stijnmoreels what do you think?
> > 
> > 
> > Aha, yeah, that could be a good alternative 👍
> 
> Applied in [`1b62d75` (#317)](https://github.com/arcus-azure/arcus.scripting/pull/317/commits/1b62d758db6566bf0a948d42b022dbcaccb49a35)

Super! 👍 Great that you find this. Shame that there isn't a `.Context` property or something like that that we could use, but in any case, this alternative gives use enough security that the API Management service really exists.
Good job!

stijnmoreels (approved):

PR #326, opened 2022-09-09
docs: added a guidelines section

pim-simons:
Added a guidelines section to include best practices, for now this only includes installation.

Closes https://github.com/arcus-azure/arcus.scripting/issues/284


docs/preview/02-Guidelines/Installation.md
```diff
@@ -0,0 +1,16 @@
...
+
+To have access to the Arcus Scripting features, you have to import the modules.
+The best practice for usage in your build and release pipelines is to use the following commands:
+
+``` powershell
+PS> Set-PSRepository -Name PSGallery -InstallationPolicy Trusted
+PS> Install-Module -Name Arcus.Scripting.{Module} -AllowClobber
+```
+
+This drastically improves performance over using the `-Force` parameter and as such, usage of the `-Force` parameter is not recommended.
```
stijnmoreels, line 16: Great to have this! 👍

stijnmoreels (approved):
💃

stijnmoreels:
Should we then remove this from the DevOps page? Or do you think that since it's mostly in DevOps scenario's that this problem occur, that we would leave it there?

pim-simons:
> Should we then remove this from the DevOps page? Or do you think that since it's mostly in DevOps scenario's that this problem occur, that we would leave it there?

I would keep it in, people can overlook the guidelines page and by having it in the DevOps page we make sure the correct way is used.

Issue #318, opened 2022-08-17
Remove table of contents in feature docs pages


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Currently, each feature documentation page has a table of contents in the beginning of the page, but this is generated by the side on the right-side in a small sidebar, causing duplication on the page.

**Describe the solution you'd like**
We should remove the table of contents in the feature docs page itself to avoid duplication and have cleaner/smaller pages.

PR #327, opened 2022-09-09
docs: removed table of contents

pim-simons:
Removed table of contents as it is already handled by the sidebar.

Closes https://github.com/arcus-azure/arcus.scripting/issues/318


stijnmoreels (approved):
Ah, that must have felt good; deleting stuff 😉 .

PR #342, opened 2022-12-02
fix: Update Algolia Search to V3

Mcalewaert:


stijnmoreels (approved):
Aha! That indeed does the trick. 🥳 Thx for investigating this.
If we know what the APP ID is, we can do the other repo's by ourselves, I think.
This value is added to the Netlify environment variables in the build & deploy tab, right?

pim-simons (approved):

PR #349, opened 2022-12-23
test: assert trimmed day in integration account tests

stijnmoreels:
Since the integration tests were asserting on seconds, we would sometimes get failues in the like of `Expected collection @('2022-12-23T05:33:25', '2022-12-23T05:33:24') to contain '2022-12-23T05:26:03', but it was not found.` because there is a time delay somewhere that causes the expected creation time of Azure Integration account types to shift. Therefore, it would be better to assert the creation time trimmed on the day instead of on the second. Then, only when you work arround midnight and run the tests at an exact time, you would maybe run into this 😀.


pim-simons (approved):
Seems like a good change to me 👍🏻

Issue #155, opened 2021-04-01
Streamline and improve module logging across PS modules


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Right now, we use solely the `Write-Host` to write diagnostic information. This is great for current session view, but we should look into a better approach that makes sure the logging isn't lost afterwards + we should come up with a more streamlined approach when we log in our PS modules.

**Describe the solution you'd like**
[`PSFramework`](https://psframework.org/documentation/documents/psframework/logging/basics/writing-messages.html) might by an option.

**Describe alternatives you've considered**
We could also start first with using more `Write-*` variants so that the output is more clearer and structured. (using `Write-Verbose` for tracing instead of `Write-Host` for example).

mbraekman:
Not sure if we should go for an additional framework for this, as it adds an additional dependency, while in 99% of the cases I expect using the `Write-Host`/`Write-Error`/...-combination will definitely suffice.
But you are correct that perhaps we should have a better look at when to use which `Write-*`-operation.

stijnmoreels:
Scheme of logging:
- Warning-like: Write-Warning "<message>"
- Trace-like: Write-Verbose "<message>"
- Success-like: Write-Host "<message>" -ForegroundColor Green
- Information-like: Write-Host "<message>"

fgheysels:
fyi,
Write-Host is awrapper around Write-Information
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-host?view=powershell-7.3

I think the above logging-scheme is OK.

PR #351, opened 2022-12-26
chore: streamline logging across modules

stijnmoreels:
Streamline logging and level across Powershell modules from `Arcus.Scripting.AzureActiveDirectory` to `Arcus.Scripting.LogicApps`, alphabetically.

Relates to #155

### DIFF
--- a/src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
+++ b/src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
@@ -38,7 +38,7 @@ try {
         $adApplication.AppRole += $newRole
 
         Update-AzADApplication -ObjectId $adApplication.Id -AppRole $adApplication.AppRole
-        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)e'"
+        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)'"
 
         $currentAppRole = $newRole
     } else {
--- a/src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
+++ b/src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
@@ -38,11 +38,11 @@ try {
         $adApplication.AppRole += $newRole
 
         Update-AzADApplication -ObjectId $adApplication.Id -AppRole $adApplication.AppRole
-        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)'" -ForegroundColor White
+        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)e'"
 
         $currentAppRole = $newRole
     } else {
-        Write-Host "Active Directory Application '$($adApplication.DisplayName)' already contains the role '$Role'" -ForegroundColor Yellow
+        Write-Warning "Active Directory Application '$($adApplication.DisplayName)' already contains the role '$Role'"
         $currentAppRole = $adApplication.AppRole | Where-Object Value -eq $Role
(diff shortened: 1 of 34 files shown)


src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
```diff
@@ -38,11 +38,11 @@ try {
         $adApplication.AppRole += $newRole
 
         Update-AzADApplication -ObjectId $adApplication.Id -AppRole $adApplication.AppRole
-        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)'" -ForegroundColor White
+        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)e'"
```
pim-simons, line 41:
```suggestion
        Write-Host "Added role '$Role' to Active Directory Application '$($adApplication.DisplayName)'"
```

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicAppsFromConfig.ps1
```diff
@@ -61,35 +61,35 @@ function ExecuteCheckType() {
...
                                 }
                             }
-                        }else{
-                            Write-Host("Found no more waiting or running runs for '$logicApp', executing stopType for Logic App")
+                        } else{
+                            Write-Verbose "Found no more waiting or running runs for Azure Logic App '$logicApp', executing stopType for Logic App"
                             ExecuteStopType -resourceGroupName $ResourceGroupName -LogicAppName $logicApp -stopType $batch.stopType
                         }                    
-                        Write-Host("> Check 'NoWaitingOrRunningRuns' executed successfully on '$logicApp'")   
+                        Write-Host "Check 'NoWaitingOrRunningRuns' executed successfully on Azure Logic APp '$logicApp'" -ForegroundColor Green
```
pim-simons, line 79:
```suggestion
                        Write-Host "Check 'NoWaitingOrRunningRuns' executed successfully on Azure Logic App '$logicApp'" -ForegroundColor Green
```

pim-simons (changes requested):
Looking good 👍🏻 
We should probably also update the docs as the logging has changed, for example https://scripting.arcus-azure.net/Features/powershell/azure-logic-apps#disable-an-azure-logic-app states:
``` powershell
PS> Disable-AzLogicApp -ResourceGroupName "rg-common-dev" -LogicAppName "rcv-shopping-order-sftp"
# Access-token and subscriptionId retrieved
# Attempting to disable rcv-shopping-order-sftp

# Successfully disabled rcv-shopping-order-sftp
```

And that is not correct anymore after this PR.

stijnmoreels:
> We should probably also update the docs as the logging has changed,

Yes, I was thinking about that. Maybe we should not have the 'exact' logging in the docs as it would be rather hard/tedious to get them up-to-date all the time. Maybe we can only include a truncated version of information-like/successful messages?

src/Arcus.Scripting.ActiveDirectory/Scripts/Add-AzADAppRoleAssignment.ps1
stijnmoreels: Woops, what happened here 😅 , thx!

pim-simons:
> > We should probably also update the docs as the logging has changed,
> 
> Yes, I was thinking about that. Maybe we should not have the 'exact' logging in the docs as it would be rather hard/tedious to get them up-to-date all the time. Maybe we can only include a truncated version of information-like/successful messages?

I get your point, but I honestly quite like the way we have it setup in the docs right now, it shows you exactly what to expect which I really like.

stijnmoreels:
Now, it is updated with the full correct information/success messages.

pim-simons (approved):
Looks good to me 🥇

PR #354, opened 2022-12-30
chore: remove all module

### LINKED ISSUE #347: Relevance of `Arcus.Scripting.All` PowerShell module
Currently, we have two `.All` PowerShell modules in our solution: 1 that collects all the Azure Storage-related modules, and 1 that collects all our modules. The problem with the second one is that it is harder to maintain and by time less relevant as there may be no scenarios that requires all our PowerShell modules.

We can remove the `Arcus.Scripting.All` PowerShell module from our solution and provide a deprecation message in the PowerShell gallery saying that it will no longer be maintained and that references to more concrete modules is necessary.

stijnmoreels:
We didn't have a lot of modules in the beginning, but now there are so many that an `.All` module encapsulating everything becomes rather tedious instead of helpful.

Closes #347


stijnmoreels:
What do you think, Pim? Could we agree on removing this?

pim-simons (approved):

pim-simons:
> What do you think, Pim? Could we agree on removing this? Or do you see other use-cases for this `.All` module?

I agree, I don't see any  use-case for this and personally never used the `.All` module either.

pim-simons:
I do see the tests failing though
`The required module 'Arcus.Scripting.Storage.Blob' is not loaded`

pim-simons (changes requested):
Tests seem to be failing

stijnmoreels:
Ah, was a bit too hasty with removing the `.All` stuff in the build script, as they are still valid for our `Arcus.Scripting.Storage.All` module. 😅 I'll change that.

pim-simons (approved):

PR #367, opened 2023-02-09
fix: active directory tag

pim-simons:
Push to Powershell Gallery is failing with this error `Response status code does not indicate success: 400 (The package is invalid. The error encountered was:'A package's Tags property extracted from the PowerShell manifest may not contain spaces: 'Active Directory'.'. Correct the error and try again.).`

Apparently whitespaces are not allowed in the tags, this PR fixes that.


stijnmoreels (approved):
Oef, nice catch! 👍

PR #375, opened 2023-02-16
chore: update w/ ps release verification

### LINKED ISSUE #368: Improve PowerShell Gallery release step by doing a verification of all published packages
It is a good think that right now, the deployment of packages is not stopped when a single package failed to upload to the PowerShell Gallery. The problem, though, is that the build does not fail at this point.

Make sure that there's a final check that verifies if all the returned HTTP status codes from the packages are `Created` and not `BadRequest`. Also, make sure that the `Conflict` is seen as a success, as it means that the package is already there.

⚠️ Make sure that this is done at the very end so as not to block the release of other packages.

It is good to know that the automatically created GitHub release is a draft release (not published) so it can easily be removed if the release pipeline is run twice.

stijnmoreels:
Attempt to verify if none of the release PowerShell modules return a `BadRequest` without halting the release of the remaining modules.

Closes #368

### DIFF
--- a/build/psgallery-release.yml
+++ b/build/psgallery-release.yml
@@ -150,7 +150,14 @@ stages:
             displayName: 'Install NuGet'
           - powershell: |
               Get-ChildItem -Path $env:ARTIFACT_DIR -Filter *.nupkg -Recurse |
-                % { & "nuget" push $_.FullName -Source $(Source) -ApiKey $(NuGet.ApiKey) -SkipDuplicate }
+                % { & "nuget" push $_.FullName -Source $(Source) -ApiKey $(NuGet.ApiKey) -SkipDuplicate | Out-File log.txt -Append }
+              
+              $log = Get-Content log.txt
+              cat $log
+              if ($log -match "BadRequest") {
+                echo "##vso[task.logissue type=error]Not all PowerShell modules were pushed correctly to the PowerShell Gallery"
+                exit 1
+              }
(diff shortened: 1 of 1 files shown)


build/psgallery-release.yml
```diff
@@ -150,7 +150,14 @@ stages:
...
               Get-ChildItem -Path $env:ARTIFACT_DIR -Filter *.nupkg -Recurse |
-                % { & "nuget" push $_.FullName -Source $(Source) -ApiKey $(NuGet.ApiKey) -SkipDuplicate }
+                % { & "nuget" push $_.FullName -Source $(Source) -ApiKey $(NuGet.ApiKey) -SkipDuplicate | Out-File log.txt -Append }
+              
+              $log = Get-Content log.txt
+              cat $log
+              if ($log -match "BadRequest") {
+                echo "##vso[task.logissue type=error]Not all PowerShell modules were pushed correctly to the PowerShell Gallery"
+                exit 1
+              }
```
stijnmoreels, line 160: We log during release, and verify at the end.
Would want to check this with a pre-release with both a invalid and valid release version, maybe. Not sure how otherwise we could test this 🤔 .

pim-simons (approved):
Seems like a good solution to me

build/psgallery-release.yml
pim-simons: Difficulty is in how to test indeed

Issue #379, opened 2023-03-24
Setting Azure DevOps variable as secret


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Currently, we provide the capability to set an Azure DevOps pipeline variable in our `Arcus.Scripting.DevOps` PowerShell module, but this functionality does not provide any switch parameter to save the variable as a 'secret' pipeline variable (value won't be shown in logs).

**Describe the solution you'd like**
Add a switch parameter to the `Set-AzDevOpsVariable` command so that we can save secret pipeline variables.

**Additional context**
Marking a pipeline variable as a secret is done by adding `issecret=true` to the outputed text:
```powershell
Write-Host "##vso[task.setvariable variable=$secretName;issecret=true]$secretValue`
```

PR #380, opened 2023-04-11
feat: Added switch to set variable as secret

pim-simons:
Adds a switch to set a variable as a secret.

Closes https://github.com/arcus-azure/arcus.scripting/issues/379

### DIFF
--- a/docs/preview/03-Features/powershell/azure-devops.md
+++ b/docs/preview/03-Features/powershell/azure-devops.md
@@ -18,16 +18,26 @@ PS> Install-Module -Name Arcus.Scripting.DevOps -Repository PSGallery -AllowClob
 
 Assign a value to a DevOps pipeline variable during the execution of this pipeline.
 
-| Parameter       | Mandatory | Description                                       |
-| --------------- | --------- | ------------------------------------------------- |
-| `Name`          | yes       | The name of the variable to set in the pipeline   |
-| `Value`         | yes       | The value of the variable to set in the pipeline  |
+| Parameter             | Mandatory | Description                                       |
+| --------------------- | --------- | ------------------------------------------------- |
+| `Name`                | yes       | The name of the variable to set in the pipeline   |
+| `Value`               | yes       | The value of the variable to set in the pipeline  |
+| `SetVariableAsSecret` | no        | The switch to set the variable as a secret        |
 
 **Example**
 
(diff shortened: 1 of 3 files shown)


pim-simons:
Looking at the code and testing the currently released version, it does not seem that this task worked correctly...
I changed `Write-Host "#vso[task.setvariable variable=$Name] $Value"` to `Write-Host "##vso[task.setvariable variable=$Name] $Value"`. Note the extra `#`, otherwise the variable is not stored in the pipeline.

stijnmoreels (approved):
Oh, so this is a bugfix and feature in one, great! 😄 
Since this only works on a build server, it is indeed a bit harder to tests locally. Maybe we could check if there is an environment variable available after running the command instead of asserting on the `Write-Host`?

docs/preview/03-Features/powershell/azure-devops.md
```diff
@@ -18,16 +18,26 @@ PS> Install-Module -Name Arcus.Scripting.DevOps -Repository PSGallery -AllowClob
...
 
-| Parameter       | Mandatory | Description                                       |
-| --------------- | --------- | ------------------------------------------------- |
-| `Name`          | yes       | The name of the variable to set in the pipeline   |
-| `Value`         | yes       | The value of the variable to set in the pipeline  |
+| Parameter             | Mandatory | Description                                       |
+| --------------------- | --------- | ------------------------------------------------- |
+| `Name`                | yes       | The name of the variable to set in the pipeline   |
+| `Value`               | yes       | The value of the variable to set in the pipeline  |
+| `SetVariableAsSecret` | no        | The switch to set the variable as a secret        |
```
stijnmoreels, line 25: Would `-AsSecret` be too abstract? Otherwise it's a bit 'duplication' of 'variable' in both function and parameter switch.

stijnmoreels: Fine by me to leave it as this, though.

pim-simons: I like `AsSecret` more, so renamed to that 👍🏻 see [`7096e07` (#380)](https://github.com/arcus-azure/arcus.scripting/pull/380/commits/7096e07d7cca0c16a901ba66fcf5fe9f441a83cb)

Issue #390, opened 2023-05-25
`Get-AzLogicAppRunHistory` command does not use the `-FollowNextPageLink` parameter


pim-simons:
**Describe the bug**
The scripts `Cancel-AzLogicAppRuns`, `Disable-AzLogicAppsFromConfig` and `Resubmit-FailedAzLogicAppRuns` use the `Get-AzLogicAppRunHistory` command but do not specify the `-FollowNextPageLink` parameter, meaning that only a subset of possible instances is returned. 

**To Reproduce**
Let 100 instances of a Logic App fail, run the `Resubmit-FailedAzLogicAppRuns` script and notice that not all failed Logic Apps have been resubmitted.

**Expected behavior**
Use the `-FollowNextPageLink` parameter on the `Get-AzLogicAppRunHistory` command.

stijnmoreels:
Aha, that's an interesting case. Good that we will have this fixed. 👍

PR #391, opened 2023-05-25
fix: Added the `-FollowNextPageLink` parameter to the `Get-AzLogicAppRunHistory` command

pim-simons:
Fixes https://github.com/arcus-azure/arcus.scripting/issues/390

### DIFF
--- a/src/Arcus.Scripting.LogicApps/Scripts/Cancel-AzLogicAppRuns.ps1
+++ b/src/Arcus.Scripting.LogicApps/Scripts/Cancel-AzLogicAppRuns.ps1
@@ -4,7 +4,7 @@ param(
 )
 
 try{
-    $runs = Get-AzLogicAppRunHistory -ResourceGroupName $ResourceGroupName -Name $LogicAppName | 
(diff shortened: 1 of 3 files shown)


src/Arcus.Scripting.LogicApps/Scripts/Cancel-AzLogicAppRuns.ps1
```diff
@@ -4,7 +4,7 @@ param(
 )
 
 try{
```
stijnmoreels, line 6:
```suggestion
try {
```

stijnmoreels (approved):
We could have run into this if we would have tried in the tests resubmitting 100 Logic Apps? That's probably not workable for our here, I imagine?
Wondering how we else can verify in our tests that it now take into account different pages. 🤔

pim-simons:
> We could have run into this if we would have tried in the tests resubmitting 100 Logic Apps? That's probably not workable for our here, I imagine? Wondering how we else can verify in our tests that it now take into account different pages. 🤔

Yeah but hard to test, I ran into this when using the script and having to resubmit thousands of Logic Apps 😉

PR #392, opened 2023-05-26
fix: use latest version of nuget during release

pim-simons:
The release to the powershell gallery sometimes failes because of the version of the nuget tool. This PR makes sure we always use the latest version of the tool.


stijnmoreels (approved):
Fingers crossed 🤞

PR #393, opened 2023-05-26
fix: fix NuGetToolInstaller task

pim-simons:
wrong notation used, fixes the NuGetToolInstaller task


stijnmoreels (approved):
Hehe, also slipped by me. 😄

PR #395, opened 2023-06-30
fix: Redundant use of Get-AzLogicAppRunHistory

ElectricCowBoy123:
Get-AzLogicAppRunHistory is used multiple times when it only needs to be used once, optimisation


fgheysels (approved):
Thanks for this; indeed a nice optimization.

pim-simons (approved):

pim-simons:
Many thanks for contributing, looks good to me 🥇 
Will merge after tests are successful.

As discussed with @ElectricCowBoy123, I think we can optimize a bit futher by introducing an optional parameter `MaximumFollowNextPageLink` with a default of for example set to 10 (which would cover most usecases I think) but I will create a separate PR for that. 

Idea is to create a fix release asap 👍🏻

Issue #255, opened 2021-12-27
Setting ARM outputs in Azure DevOps variable group integration test fails


stijnmoreels:
**Describe the bug**
The integration test of setting the ARM outputs in an Azure DevOps variable group only works when triggered manually; it fails when it's triggered automatically by GitHub.

**To Reproduce**
Unskip test and create PR.

**Expected behavior**
All should be green.

**Additional context**
 - v0.5, current version on `master` since #252

PR #400, opened 2023-07-06
fix: enable integration test for devops variable group description

pim-simons:
Enable integration test for devops variable group description.
We use a different way to authenticate towards the DevOps API since the `$env:SYSTEM_ACCESSTOKEN` doesn't work for variable groups when triggered by GitHub.

Closes https://github.com/arcus-azure/arcus.scripting/issues/255


stijnmoreels (approved):
Thx for this!

Issue #405, opened 2023-07-12
Integration account integration tests sometimes fail with 'Not Found'


stijnmoreels:
**Describe the bug**
Sometimes the integration tests related to the Azure Integration Account fail with a 'Not Found' error.

```
[-] Arcus Azure Integration Account integration tests.Uploading Schemas into an Azure Integration Account.Create a single schema in an Integration Account succeeds 3.04s (3.03s|3ms)
 CloudException: Operation returned an invalid status code 'NotFound'
 at UploadSchema, /home/vsts/work/1/s/src/Arcus.Scripting.IntegrationAccount/Scripts/Set-AzIntegrationAccountSchemas.ps1:39
 at <ScriptBlock>, /home/vsts/work/1/s/src/Arcus.Scripting.IntegrationAccount/Scripts/Set-AzIntegrationAccountSchemas.ps1:73
 at Set-AzIntegrationAccountSchemas, /home/vsts/work/1/s/src/Arcus.Scripting.IntegrationAccount/Arcus.Scripting.IntegrationAccount.psm1:39
 at <ScriptBlock>, /home/vsts/work/1/s/src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.IntegrationAccount.tests.ps1:44
Schema 'NestedSchema.xsd' has been uploaded into the Azure Integration Account 'arcus-scripting-dev-ne-ia'
```

**To Reproduce**
Run CI pipeline.

**Expected behavior**
Azure Integration Account integration tests succeed.

**Additional context**
 - v1.1.0

PR #406, opened 2023-07-12
feat: add retries to integration account actions

pim-simons:
Adds retries to integration account actions as these seem to fail regularly since last week, it seems like connection issues.

Closes https://github.com/arcus-azure/arcus.scripting/issues/405


stijnmoreels (approved):
Let's hope this works.🤞
Thx!

PR #411, opened 2023-08-24
feat: Use separate pipeline tasks for tests

pim-simons:
In the past months we have had some issues with the integration tests taking a very long time and sometimes errorring out, you can restart the failed pipeline task but the whole process starts to take a very long time. 

This is a first attempt to splitting up the various tests into separate pipeline tasks so they can run in parallel and they can be restarted separately.


build/psgallery-release.yml
```diff
@@ -69,14 +103,16 @@ stages:
...
+              testName: '$(Project).${{UnitTest.name}}'
 
   - stage: IntegrationTests
     displayName: Integration Tests
     dependsOn: Build
     condition: succeeded()
     jobs:
-      - job: RunIntegrationTests
-        displayName: 'Run integration tests'
+    - ${{ each IntegrationTest in parameters.IntegrationTests }}:
```
stijnmoreels, line 113: Maybe an idea to make a job template from the `run-pester-tests.yml` so we only have to specify the loop once.

build/psgallery-release.yml
```diff
@@ -10,6 +10,39 @@ parameters:
     displayName: 'Prerelease string (ex. -alpha, -alpha1, -BETA, -update20171020) or none'
     type: 'string'
     default: 'none'
+  - name: 'UnitTests'
+    type: object
+    default: 
+    - name: ActiveDirectory
```
stijnmoreels, line 16: Maybe we can generate this list based on the test files, that way, we only have to pass in the test folder name to the `run-pester-tests.yml`.

stijnmoreels (commented):
Added some suggestions to make it more maintainable. Otherwise, good job! This will make it a bit more easier to figure out failures.
If I can book my time more on Arcus, I can also look for making the tests a bit more resilient.

build/psgallery-release.yml
pim-simons: I don't think that will work since the list of unit tests is not the same as the list of integration tests. 
For example `ApiManagement` does not have integration tests.

build/psgallery-release.yml
pim-simons: I had the same idea initially but after some searching found that apparently this is not possible, see for example https://stackoverflow.com/questions/73332856/azure-pipelines-iterate-through-files

stijnmoreels: Hmm, that StackOverflow issue talks about you can't use absolute paths, if I'm correct. We've done in the past these kinds of things before, if we use `$(Build.SourcesDirectory)` it would be alright. 
Just thinking that a hard-coded list would be rather troublesome to maintain, especially if things change.

pim-simons: Will do some more research into this 👍🏻

stijnmoreels (approved):
Great! Thx.

Issue #413, opened 2023-10-18
logic app standard support for Azure Logic Apps script


LaurentAerens:
**Is your feature request related to a problem? Please describe.**
Currently we are using logic apps standard more and more and they have workflows instead of logic app names

**Describe the solution you'd like**
I would like it is the current logic app scripts (especially Disabling Azure Logic Apps from configuration file) would have support for disabling and enabling workflows in standard logic apps.

**Describe alternatives you've considered**
Currently we disable and enable them by manipulating the app-settings:

```
   {
    "name": "Workflows.testflow-rcv.FlowState",
    "value": "Disabled"
  },
```

stijnmoreels:
Great, thx @LaurentAerens for the issue! Seems like it would be a great addition to keep our Scripting repo up-to-date.

pim-simons:
@LaurentAerens I have been looking into this but unfortunately there is no API to disable or enable a Logic App workflow, see the docs here https://learn.microsoft.com/en-us/rest/api/appservice/workflows.

I checked what kind of HTTP call the browser does when enabling/disabling a Logic App Standard workflow and it basically executes a PUT to add an app setting indicating the workflow is disabled (the same way you do it now). 

We could leverage [this script](https://scripting.arcus-azure.net/Features/powershell/azure-app-service#set-an-app-setting-within-an-azure-app-service) and see if we can work with that as this script basically does what we need to enable/disable a workflow.

@LaurentAerens does this approach sound OK to you?

pim-simons:
Already did a quick test and this approach works fine, see https://github.com/arcus-azure/arcus.scripting/pull/415/files

LaurentAerens:
sorry for the late response, Didn't get any update emails from this issue.

Currently we also the app setting in the pipeline to disable the logic apps but is this could be done with a arcus script that would help us a lot.

pim-simons:
> sorry for the late response, Didn't get any update emails from this issue.
> 
> Currently we also the app setting in the pipeline to disable the logic apps but is this could be done with a arcus script that would help us a lot.

Sounds good :) I will pick this up further.
[...]

PR #415, opened 2023-10-20
feat: Support Logic App Standard

pim-simons:
Add support for Logic App Standard in our Logic Apps scripts. 

Closes https://github.com/arcus-azure/arcus.scripting/issues/413


stijnmoreels:
> I couldn't find a way to implement the prefix stuff we discussed [here](https://github.com/arcus-azure/arcus.scripting/issues/413#issuecomment-1772823385), since this works really a lot different in Logic App Standard. Might be worth a look later on as an enhancement though.
> 
> I split up the documentation pages to `Logic App - Consumption` and `Logic App - Standard` as I feel this makes the docs much clearer.

Yes, okay, a big thx for investigating in this and good thinking in splitting this up.
Should we add some redirects to the https://github.com/arcus-azure/arcus.scripting/blob/main/netlify.toml for that best?

docs/preview/03-Features/powershell/azure-logic-apps-consumption.md
```diff
@@ -73,20 +73,20 @@ PS> Resubmit-FailedAzLogicAppRuns `
...
-|                   |           | When not provided, it will be retrieved from the current context (Get-AzContext).                                   |
-| ResourceGroupName | yes       | The resource group containing the Azure Logic Apps.                                                                 |
-| LogicAppName      | yes       | The name of the Azure Logic App to be disabled.                                                                     |
-| ApiVersion        | no        | The version of the management API to be used.  (default: `2016-06-01`)                                              |
-| AccessToken       | no        | The access token to be used to disable the Azure Logic App.                                                         |
-|                   |           | When not provided, it will be retrieved from the current context (Get-AzContext).                                   |
+| Parameter           | Mandatory | Description                                                                                                         |
```
stijnmoreels, line 76: Hehe, thx for fixing the table ;)

docs/preview/03-Features/powershell/azure-logic-apps-consumption.md
```diff
@@ -149,13 +149,13 @@ Typically done the first task of the release pipeline, right before the deployme
...
-| EnvironmentName   | no        | The name of the Azure environment where the Azure Logic App resides. (default: `AzureCloud`)                                                                |
-| ApiVersion        | no        | The version of the management API to be used.  (default: `2016-06-01`)                                                                                      |
+| Parameter           | Mandatory | Description                                                                                                                                                 |
+| ------------------- | --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ResourceGroupName` | yes       | The resource group containing the Azure Logic Apps.                                                                                                         |
+| `DeployFileName`    | yes       | If your solution consists of multiple interfaces, you can specify the flow-specific name of the configuration file.                                         |
```
stijnmoreels, line 155: Was this `DeployFileName` set to mandatory something wrong with the docs, or is this a breaking change?

docs/preview/03-Features/powershell/azure-logic-apps-standard.md
```diff
@@ -0,0 +1,354 @@
...
+
+| Parameter                   | Mandatory | Description                                                                                                                                         |
+| --------------------------- | --------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `EnvironmentName`           | no        | The name of the Azure environment where the Azure Logic App resides. (default: `AzureCloud`)                                                        |
+| `ResourceGroupName`         | yes       | The resource group containing the Azure Logic App.                                                                                                  |
+| `LogicAppName`              | yes       | The name of the Azure Logic App containing the workflow for which the runs will be cancelled.                                                       |
+| `WorkflowName`              | yes       | The name of the workflow within the Azure Logic App for which the runs will be cancelled.                                                           |
```
stijnmoreels, line 25: The `WorkflowName` was also optional, no?

docs/preview/03-Features/powershell/azure-logic-apps-standard.md
```diff
@@ -0,0 +1,354 @@
+---
+title: "Azure Logic Apps - Standard"
+layout: default
+---
+
+# Azure Logic Apps - Standard
```
stijnmoreels, line 6: Just to check: this file is almost a copy/paste of the other one, except that it allows for additional parameters?

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicApp.ps1
```diff
@@ -3,35 +3,42 @@ param(
...
     [Parameter(Mandatory = $false)][string] $ApiVersion = "2016-06-01",
     [Parameter(Mandatory = $false)][string] $AccessToken = ""
 )
 
 try{
-    if($SubscriptionId -eq "" -or $AccessToken -eq ""){
-        # Request accessToken in case the script contains records
-        $token = Get-AzCachedAccessToken
+    if ($WorkflowName -eq "") {
+        if ($SubscriptionId -eq "" -or $AccessToken -eq ""){
```
stijnmoreels, line 13:
```suggestion
        if ($SubscriptionId -eq "" -or $AccessToken -eq "") {
```
Nitpicking all the way. :D

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicAppsFromConfig.ps1
```diff
@@ -14,27 +14,31 @@ function ExecuteStopType() {
     param
```
stijnmoreels, line 14: This file is starting to get complex. We might refactor some inner functionality to separate functions, maybe? Can be looked at later, of course.

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -26,15 +24,13 @@ InModuleScope Arcus.Scripting.LogicApps {
                     }
                 }
 
-                # Act
-                Enable-AzLogicApp -ResourceGroupName $resourceGroupName -LogicAppName $logicAppName
-
+                # Act
```
stijnmoreels, line 27:
```suggestion
                # Act
```

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -196,12 +200,22 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+                    if ($script:i -gt 2) {
+                        Write-Host "Returning empty (no running, no waiting) runs"
+                        return @()
+                    } else {
+                        Write-Host "Returning 1 waiting run"
+                        return @(
+                            [pscustomobject]@{ Status = "Waiting" }
+                        )
+                    }
                 }
```
stijnmoreels, line 219: Yeah, this is probably the best we can do, with the index. Maybe there's some extension on the Pester library we can write to create 'sequenced stubs' or something.

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -678,4 +710,467 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+                    } else {
+                        Write-Host "Returning 1 running run"
+                        return [pscustomobject] @{
+                            Content = '{"value":[{"properties":{"status":"Running"}}]}'
+                        }
+                    }
+                }
+                Mock Disable-AzLogicApp {}
+                Mock Get-AzCachedAccessToken -MockWith {
+                    $Global:subscriptionId = '123456'
```
stijnmoreels, line 904: Thought that those global variables were for testing... It feels a bit an anti-pattern to have them there as global ones to use them here. Thinking maybe even a security risk that other following scripts could also access them? Isn't there another way we can stub out some function and return them? Or maybe create `$script:` variables here and store them once this is called?

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -678,4 +710,467 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+
+                # Act
+                Disable-AzLogicAppsFromConfig -DeployFileName "$PSScriptRoot\Files\LogicApps\Standard\deploy-orderControl-noWaitingOrRunningRunsWithImmediate.json" -ResourceGroupName $resourceGroup
+
+                # Assert
+                Assert-MockCalled Get-AzCachedAccessToken -Scope It -Times 1
+                Assert-MockCalled Invoke-WebRequest -Scope It -Exactly 12
+                Assert-MockCalled Disable-AzLogicApp -Scope It -Exactly 5 -ParameterFilter { $resourceGroupName -eq $resourceGroup }
+            }
+            It "Disables all logic apps when checkType = NoWaitingOrRunningRuns with found waiting and running runs for stopType = Immediate" {
```
stijnmoreels, line 996: Think these tests are all the same like we already had, right? If not, could you point me the new ones? Thx.

stijnmoreels (changes requested):
Some questions, some suggestions, some changes. I mostly trust your knowledge on this. The Logic Apps scripting has become rather big and complex. It might be useful to refactor some common functionality out and place certain inner logic in separate functions within the script file(s). But can be looked another time. I believe most changes are the same for consumption/standard?

And of course, a BIG thank you for your hard work! 🎖️

docs/preview/03-Features/powershell/azure-logic-apps-consumption.md
pim-simons: Something wrong with the docs, the `DeployFileName` should be mandatory 😄

docs/preview/03-Features/powershell/azure-logic-apps-standard.md
pim-simons: It is mostly the same indeed, but other parameters are required or not relevant.
I split them up between consumption and standard because otherwise the docs became unclear (a lot of text, multiple examples, etc)

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicAppsFromConfig.ps1
pim-simons: Would leave it for a later enhancement indeed 👍🏻

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: You'll probably have to help me on that one, my Pester knowledge is not that extensive 😆

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: For Standard I implemented the same tests as Consumption.
Consumption tests are here: https://github.com/pim-simons/arcus.scripting/blob/feature/logic-app-standard/src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1#L5
Standard tests are here: https://github.com/pim-simons/arcus.scripting/blob/feature/logic-app-standard/src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1#L713

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: Not entirely sure what you mean, but this is only used on the tests. 
This was really the only way I could get the global variable populated if I remember correctly. But if there is a better solution I am all for it 👍🏻

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicApp.ps1
pim-simons: Loving the nitpicking 👍🏻

pim-simons:
> > I couldn't find a way to implement the prefix stuff we discussed [here](https://github.com/arcus-azure/arcus.scripting/issues/413#issuecomment-1772823385), since this works really a lot different in Logic App Standard. Might be worth a look later on as an enhancement though.
> > I split up the documentation pages to `Logic App - Consumption` and `Logic App - Standard` as I feel this makes the docs much clearer.
> 
> Yes, okay, a big thx for investigating in this and good thinking in splitting this up. Should we add some redirects to the [`main`/netlify.toml](https://github.com/arcus-azure/arcus.scripting/blob/main/netlify.toml?rgh-link-date=2023-11-17T11%3A15%3A04Z) for that best?

I honestly don't really have any knowledge on that toml file, not really sure what it does 😇 
I notice that Logic App isn't in the toml file currently?

stijnmoreels:
> I honestly don't really have any knowledge on that toml file, not really sure what it does 😇 I notice that Logic App isn't in the toml file currently?

That's right. I would believe that we need to add some `[redirects]` at the end that redirects the previous logic apps version to the standard or consumption now (whatever you find more appropriate and/or recommended to use).

I think these two should be added:
```toml
[[redirects]]
  from = "/:version/features/powershell/azure-logic-apps"
  to = "/:version/features/powershell/azure-logic-apps-standard"

[[redirects]]
  from = "/features/powershell/azure-logic-apps"
  to = "/features/powershell/azure-logic-apps-standard"
```

But I also think that we should only add them when we have a new version of the documentation released, not when they are still in `/preview`. So I would suggest to create a task for the next version so that we take this up when we release the next version.

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
stijnmoreels: Yes, my concern is that this was only used for the tests, and that it could potentially be vulnerable. I'm wondering why we need it, though, as the `Get-AzCachedAccessToken` function already returns the same information? And once it is set, it is not used for assertion, too.

src/Arcus.Scripting.LogicApps/Scripts/Cancel-AzLogicAppRuns.ps1
```diff
@@ -1,19 +1,82 @@
...
+        $runs = Get-AzLogicAppRunHistory -ResourceGroupName $ResourceGroupName -Name $LogicAppName -FollowNextPageLink -MaximumFollowNextPageLink $MaximumFollowNextPageLink | 
+            Where-Object {$_.Status -eq 'Running'}
 
-    foreach ($run in $runs) {
-        Stop-AzLogicAppRun -ResourceGroupName $ResourceGroupName -Name $LogicAppName -RunName $run.Name -Force
-        Write-Verbose "Cancelled run $run.Name for the Azure Logic App '$LogicAppName' in resource group '$ResourceGroupName'"
-    }
+        foreach ($run in $runs) {
+            Stop-AzLogicAppRun -ResourceGroupName $ResourceGroupName -Name $LogicAppName -RunName $run.Name -Force
+            Write-Verbose "Cancelled run $run.Name for the Azure Logic App '$LogicAppName' in resource group '$ResourceGroupName'"
```
stijnmoreels, line 19: I think that we need to use `$($run.Name)` to concatenate a property directly in a `string`, no?

docs/preview/03-Features/powershell/azure-logic-apps-standard.md
stijnmoreels: Ah, now I see it. This parameter is also responsible if the implementation is using consumption/standard, right?
Just a thought: this is a bit implicit, it might be a good idea to include `Standard`/`Consumption` in the function names? If yes: then we should make sure that we write an deprecated error when the 'old' ones, without the type, are still used.

src/Arcus.Scripting.LogicApps/Scripts/Cancel-AzLogicAppRuns.ps1
pim-simons: aligned the logging in https://github.com/arcus-azure/arcus.scripting/pull/415/commits/f901fcd105ec31ffd9944f338e7e9ba730abbcdb

pim-simons:
> Ah, now I see it. This parameter is also responsible if the implementation is using consumption/standard, right?
Just a thought: this is a bit implicit, it might be a good idea to include Standard/Consumption in the function names? If yes: then we should make sure that we write an deprecated error when the 'old' ones, without the type, are still used.

Yes the `WorkflowName` parameter determines if the script executes the Consumption or Standard functionality since a workflow only exists in the Logic App Standard. It is a bit implicit, but I did not want to change the names of the scripts as that would break current implementations.

pim-simons:
> Yes, my concern is that this was only used for the tests, and that it could potentially be vulnerable. I'm wondering why we need it, though, as the Get-AzCachedAccessToken function already returns the same information? And once it is set, it is not used for assertion, too.

Just tried again and without the `$Global:subscriptionId = '123456'` the test fails with:
> Cannot bind argument to parameter 'SubscriptionId' because it is an empty string.

So I honestly don't really know how else to solve this?

stijnmoreels:
> Yes the `WorkflowName` parameter determines if the script executes the Consumption or Standard functionality since a workflow only exists in the Logic App Standard. It is a bit implicit, but I did not want to change the names of the scripts as that would break current implementations.

Yes, that's why I suggested to maybe log a warning for the current module functions and create additional ones with `Consumption`/`Standard`? Can be in a follow-up PR.

pim-simons:
Integration tests fail due to https://github.com/arcus-azure/arcus.scripting/issues/416, will look into that asap.

pim-simons:
> > Yes the `WorkflowName` parameter determines if the script executes the Consumption or Standard functionality since a workflow only exists in the Logic App Standard. It is a bit implicit, but I did not want to change the names of the scripts as that would break current implementations.
> 
> Yes, that's why I suggested to maybe log a warning for the current module functions and create additional ones with `Consumption`/`Standard`? Can be in a follow-up PR.

Lets do that in a follow up indeed, than we have a bit more time to look at this 👍🏻

pim-simons:
> If I read this correctly, we retrieve the $token from the Get-AzCachedAccessToken and pass that value to the next function. Why there would be a 'global' variable accessible outside the script, I don't understand directly?

I don't really know the background of this script. 

What I think happens is that because we Mock the `Get-AzCachedAccessToken` the global variables are not set. 
That is why the `$Global:subscriptionId = '123456'` is needed. 
Does that make sense?

pim-simons:
> > If I read this correctly, we retrieve the $token from the Get-AzCachedAccessToken and pass that value to the next function. Why there would be a 'global' variable accessible outside the script, I don't understand directly?
> 
> I don't really know the background of this script.
> 
> What I think happens is that because we Mock the `Get-AzCachedAccessToken` the global variables are not set. That is why the `$Global:subscriptionId = '123456'` is needed. Does that make sense?

As discussed, I removed the global variables from the scripts where they were not already present.

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicAppsFromConfig.ps1
```diff
@@ -53,59 +57,137 @@ function ExecuteCheckType() {
...
-                                $RunningRunsCount = ($runHistory | Where-Object { $_.Status -eq "Running" }).Count
-                                $WaitingRunsCount = ($runHistory | Where-Object { $_.Status -eq "Waiting" }).Count
-                                if ($RunningRunsCount -eq 0 -and $WaitingRunsCount -eq 0) {
-                                    Write-Verbose "Found no more waiting or running runs for Azure Logic App '$logicApp', executing stopType for Logic App"
-                                    ExecuteStopType -resourceGroupName $ResourceGroupName -LogicAppName $logicApp -stopType $batch.stopType
+                        if ($_.workflows.Length -gt 0 ) {
+                            $_.workflows | ForEach-Object {
+                                $WorkflowName = $_;
+
+                                $listRunningUrl = . $PSScriptRoot\Get-AzLogicAppStandardResourceManagementUrl.ps1 -EnvironmentName $EnvironmentName -SubscriptionId $Global:subscriptionId -ResourceGroupName $ResourceGroupName -LogicAppName $LogicAppName -WorkflowName $WorkflowName -Action 'listRunning'
```
stijnmoreels, line 69: Think this script still uses the `-Global...` stuff?

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -678,4 +710,462 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+                    } else {
+                        Write-Host "Returning 1 waiting run"
+                        return [pscustomobject] @{
+                            Content = '{"value":[{"properties":{"status":"Waiting"}}]}'
+                        }
+                    }
+                }
+                Mock Disable-AzLogicApp {}
+                Mock Get-AzCachedAccessToken -MockWith {
+                    $Global:subscriptionId = '123456'
```
stijnmoreels, line 941: This can now be removed?

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -678,4 +710,462 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+                            Content = '{"value":[{"properties":{"status":"Waiting"}}]}'
+                        }
+                    }
+                }
+                Mock Disable-AzLogicApp {
+                    $ResourceGroupName | Should -Be $resourceGroup
+                    $LogicAppName | Should -BeIn $logicAppNames
+                }
+                Mock Get-AzCachedAccessToken -MockWith {
+                    $Global:subscriptionId = '123456'
```
stijnmoreels, line 981: This can now be removed?

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
```diff
@@ -678,4 +710,462 @@ InModuleScope Arcus.Scripting.LogicApps {
...
+                            Content = '{"value":[{"properties":{"status":"Waiting"}},{"properties":{"status":"Running"}}]}'
+                        }
+                    }
+                }
+                Mock Disable-AzLogicApp {
+                    $ResourceGroupName | Should -Be $resourceGroup
+                    $LogicAppName | Should -Be "snd-async"
+                }
+                Mock Get-AzCachedAccessToken -MockWith {
+                    $Global:subscriptionId = '123456'
```
stijnmoreels, line 1020: This can now be removed?

stijnmoreels (approved):
If we have refactored the script to something smaller, it will be easier to see if we have handled all cases. Thx for the work!

Created these follow-up issues:
https://github.com/arcus-azure/arcus.scripting/issues/418
https://github.com/arcus-azure/arcus.scripting/issues/419

Feel free to correct my mistakes in describing them 😉 .

src/Arcus.Scripting.LogicApps/Scripts/Disable-AzLogicAppsFromConfig.ps1
pim-simons: As discussed, this script already used the global settings, for backwards compatibility reasons we will keep this in for now and in the future remove it 👍🏻

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: Not yet, needed because this script uses the global variables. We will keep this in for now and in the future remove it 👍🏻

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: Not yet, needed because this script uses the global variables. We will keep this in for now and in the future remove it 👍🏻

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.LogicApps.tests.ps1
pim-simons: Not yet, needed because this script uses the global variables. We will keep this in for now and in the future remove it 👍🏻

docs/preview/03-Features/powershell/azure-logic-apps-standard.md
stijnmoreels: Added follow-up issue for this here: #419

Issue #426, opened 2024-05-03
`Backup-AzApiManagementService` is not working anymore


pim-simons:
**Describe the bug**
Using the `Backup-AzApiManagementService` results in 
> Operation returned an invalid status code 'BadRequest'

This seems to be an issue in the Azure Powershell code, see also https://github.com/Azure/azure-powershell/issues/23411
The solution seems to be to use managed identity to backup APIM to storage account.

**To Reproduce**
- `Backup-AzApiManagementService`

**Expected behavior**
A successful backup made

PR #427, opened 2024-05-03
fix: `Backup-AzApiManagementService`: Use managed identity for connection between APIM and Storage Account

pim-simons:
Closes https://github.com/arcus-azure/arcus.scripting/issues/426

Also had to add a delay to `Create-AzTableStorageAccountTable.ps1` as this was failing the integration tests and documentation states that a remove action can take 40 seconds: https://learn.microsoft.com/en-us/rest/api/storageservices/Delete-Table?redirectedfrom=MSDN#remarks
> Note that deleting a table is likely to take at least 40 seconds to complete. If you attempt an operation against the table while it's being deleted, the service returns status code 409 (Conflict).

### DIFF
--- a/src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
+++ b/src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
@@ -31,62 +31,30 @@ if ($storageKeys -eq $null -or $storageKeys.count -eq 0) {
     Write-Verbose "Start backing up Azure API Management instance '$($ServiceName)' in resource group '$($ResourceGroupName)'..."
     if ($BlobName -ne $null -and $BlobName -ne "") {
         if ($PassThru) {
-            if ($DefaultProfile -ne $null) {
-                if ($AccessType -eq 'UserAssignedManagedIdentity') {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru -DefaultProfile $DefaultProfile
-                } else {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru -DefaultProfile $DefaultProfile
-                }
+            if ($AccessType -eq 'UserAssignedManagedIdentity') {
+                Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru -DefaultProfile $DefaultProfile
             } else {
-                if ($AccessType -eq 'UserAssignedManagedIdentity') {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru
-                } else {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru
-                }
+                Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru -DefaultProfile $DefaultProfile
             }
         } else {
-            if ($DefaultProfile -ne $null) {
-                if ($AccessType -eq 'UserAssignedManagedIdentity') {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -DefaultProfile $DefaultProfile
-                } else {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -DefaultProfile $DefaultProfile
-                }
+            if ($AccessType -eq 'UserAssignedManagedIdentity') {
+                Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -DefaultProfile $DefaultProfile
             } else {
-                if ($AccessType -eq 'UserAssignedManagedIdentity') {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName
-                } else {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName
-                }
+                Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -DefaultProfile $DefaultProfile
             }
         }
     } else {
         if ($PassThru) {
-            if ($DefaultProfile -ne $null) {
-                if ($AccessType -eq 'UserAssignedManagedIdentity') {
-                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -PassThru -DefaultProfile $DefaultProfile
-                } else {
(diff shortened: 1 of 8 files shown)


src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
```diff
@@ -23,32 +29,64 @@ if ($storageKeys -eq $null -or $storageKeys.count -eq 0) {
...
             } else {
-                Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru
+                if ($AccessType -eq 'UserAssignedManagedIdentity') {
+                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -IdentityClientId $IdentityClientId -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru
+                } else {
+                    Backup-AzApiManagement -ResourceGroupName $ResourceGroupName -Name $ServiceName -AccessType $AccessType -StorageContext $storageContext -TargetContainerName $ContainerName -TargetBlobName $BlobName -PassThru
+                }
             }
         } else {
             if ($DefaultProfile -ne $null) {
```
stijnmoreels, line 48: I think that this was implemented a bit harder than need be. I believe that we can pass `$DefaultProfile = $null` to the `Bckup-AzApiMnagement`. Best that we do some refactoring here to make it more simple instead of 4 nested `if`'s.

src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1
```diff
@@ -47,6 +47,7 @@ if ($TableName -in $tables.Name) {
     if ($Recreate) {
         Write-Verbose "Deleting existing Azure storage table '$TableName' in the Azure storage account '$StorageAccountName'..."
         $isRemoved = Remove-AzStorageTable -Name $TableName -Context $storageAccount.Context -Force
+        Start-Sleep -Seconds 40
```
stijnmoreels, line 50: Should we add this here? Or did you forgot to remove it? Best that we not fix or enforce this in our scripts, no?

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Security.tests.ps1
```diff
@@ -57,9 +57,13 @@ InModuleScope Arcus.Scripting.Security {
                     -LockLevel CanNotDelete `
                     -Force
 
+                Start-Sleep -Seconds 10
```
stijnmoreels, line 60: Why was these delays added? I think we have another function somewhere that polls for stuff, instead of a fixed wait time, if some resource is slow.

src/Arcus.Scripting.Tests.Unit/Arcus.Scripting.ApiManagement.tests.ps1
```diff
@@ -143,14 +150,55 @@ InModuleScope Arcus.Scripting.ApiManagement {
...
+                    $AccessType | Should -Be $accessType
+                    $IdentityClientId | Should -Be $identityClientId
+                    $StorageContext | Should -be $expectedStorageContext
+                    $TargetContainerName | Should -Be $targetContainerName
+                    $TargetBlobName | Should -BeNullOrEmpty 
+                    $PassThru | Should -Be $false
+                    $DefaultProfile | Should -Be $null }
+
+                # Act
+                Backup-AzApiManagementService -ResourceGroupName $resourceGroup -StorageAccountResourceGroup $storageAccountResourceGroup -StorageAccountName $expectedStorageAccountName -ServiceName $serviceName -ContainerName $targetContainerName -AccessType $accessType -IdentityClientId $identityClientId
```
stijnmoreels, line 201: We can't integration test this yet bc it is too expensive, right? Shame, then this is indeed best we can do. Thx!

pim-simons: Indeed, bit too expensive

src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1
pim-simons: Don't really know where else to fix this.

The integration tests were consistently failing here, the docs state that a remove action can take 40 seconds: [learn.microsoft.com/en-us/rest/api/storageservices/Delete-Table?redirectedfrom=MSDN#remarks](https://learn.microsoft.com/en-us/rest/api/storageservices/Delete-Table?redirectedfrom=MSDN#remarks)

> Note that deleting a table is likely to take at least 40 seconds to complete. If you attempt an operation against the table while it's being deleted, the service returns status code 409 (Conflict).

So I added a 40 second wait time and the integration test now works.
Other option could be to maybe increase the default max wait timeout? https://github.com/pim-simons/arcus.scripting/blob/7fea36b34fa0f46334c589b40ba16fe6ce1f3f45/src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1#L7

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Security.tests.ps1
stijnmoreels: Don't know by heart, something with `Poll`, I believe. Shame that we had to add these, maybe we can investigate later on how to improve it.

src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1
stijnmoreels: Ok, wow, 40s is then indeed something that could happen; but we should maybe make this more smart by polling during that time to see if it was created, instead on hard-coding this?

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Security.tests.ps1
pim-simons: Indeed, lets pick this up later on👍🏻

src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1
pim-simons: We could increase the default `MaxRetryCount` (https://github.com/pim-simons/arcus.scripting/blob/7fea36b34fa0f46334c589b40ba16fe6ce1f3f45/src/Arcus.Scripting.Storage.Table/Scripts/Create-AzTableStorageAccountTable.ps1#L7) to `15` or `20`, that should (in combination with default of 5 seconds for the `RetryIntervalSeconds`) fix it as well.

src/Arcus.Scripting.Tests.Integration/Arcus.Scripting.Security.tests.ps1
stijnmoreels: Can you create an issue for this an link it here? Thx a lot!

src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
pim-simons: Changed in [`356a092` (#427)](https://github.com/arcus-azure/arcus.scripting/pull/427/commits/356a092f0ef1b33522e67922ca6ad4724f09dd83)

stijnmoreels: It's still a lot of `if`'s 😜 Can't we use like PowerShell's argument-collection feature to pass the arguments instead of defining them ourselves? Might simplify things. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_splatting?view=powershell-7.4

pim-simons: Good one, updated the script, still some if statements but quite a bit simpler I think: [`a6dad91` (#427)](https://github.com/arcus-azure/arcus.scripting/pull/427/commits/a6dad91c5a4561b95293114c89ad2bcd9e33c636)

stijnmoreels: Yes, like this very much, much more reaable. Thanks a lot!!

src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
```diff
@@ -2,13 +2,19 @@ param(
     [Parameter(Mandatory = $true)][string] $ResourceGroupName = $(throw "Resource group name is required"),
     [Parameter(Mandatory = $true)][string] $StorageAccountResourceGroupName = $(throw = "Resource group for storage account is required"),
     [Parameter(Mandatory = $true)][string] $StorageAccountName = $(throw "Storage account name is required"),
-    [Parameter(Mandatory = $true)][string] $ServiceName = $(throw "API managgement service name is required"),
+    [Parameter(Mandatory = $true)][string] $ServiceName = $(throw "API management service name is required"),
     [Parameter(Mandatory = $true)][string] $ContainerName = $(throw "Name of the target blob container is required"),
-    [Parameter(Mandatory = $false)][string]  $BlobName = $null,
+    [Parameter(Mandatory = $true)][string][ValidateSet('SystemAssignedManagedIdentity', 'UserAssignedManagedIdentity')] $AccessType = $(throw "The access type is required"),
```
stijnmoreels, line 7: Comment is a bit late to the party 😜 I was wondering if we could change the parameter type here to something like an enumeration, so that we could o `-AccessType SystemAssignedManagedIdentity`. I think the intelisense will also then be able to help, I believe.

stijnmoreels (approved):
Great! Thx for keeping this repo up to date with much-needed updates. 💯

src/Arcus.Scripting.ApiManagement/Scripts/Backup-AzApiManagementService.ps1
pim-simons: Better late than never 😉 

I'm not really sure what having an enum would provide above having the `ValidateSet`, as I understand this already has tab completion: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_argument_completion?view=powershell-7.4#validateset-attribute.
> The ValidateSet attribute specifies a set of valid values for a parameter or variable and enables tab completion. 

We also use the `ValidateSet` in several other scripts, might be better to keep this all in line? 
What do you think?

stijnmoreels: Aha, okay, thx for the context. Yes, then it is better to use this approach. 👍

PR #437, opened 2024-11-15
feat: set az devops group var

### LINKED ISSUE #435: ✨ Add `Set-AzDevOpsGroupVariable` for more fine-grained control of the DevOps variable group
We currently have a `Set-AzDevOpsArmOutputsToVariableGroup` that provides a way to place all output variables of a ARM/Bicep deployment to an Azure DevOps variable group. But for a more fine-grained control on what needs to be set - and to also set variables that aren't always outputed, there is no easy way to add such a variable.

Adds an accompanied `Set-AzDevOpsGroupVariable` that allows us to set a single variable in an Azure DevOps variable group.

**Describe alternatives you've considered**
People are forced to output everything they need, even when that information is not applicable or related to the ARM/Bicep's components. That's how it's currently done/available to do.

**Additional context**
In the `Arcus.Scripting.DevOps` module.

stijnmoreels:
Make sure that users can update DevOps variable groups independantly from any previous deployments.

Closes #435


fgheysels (approved):

stijnmoreels:
Will be closing this as it is not in line with what the DevOps Practice will recommend.

Issue #436, opened 2024-11-14
📦 Upgrade `Az.*` dependent modules


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
We currently depend on `Az.*` packages that are 4 or more years older. This could lead to security, conflict/resolve issues and other problems.

**Describe the solution you'd like**
Best that we upgrade all the dependend packages and maybe do a major update if there are breaking changes.

PR #439, opened 2024-11-28
feat: Upgrade Az.* dependent modules 

pim-simons:
Closes https://github.com/arcus-azure/arcus.scripting/issues/436


stijnmoreels:
Are these changes too big to be in a minor release? Meaning: would releasing this as the next minor release break any scripts on existing DevOps piprlines that use it?
Wondering if we should play it safe and see this as a breaking change. 🤔

pim-simons:
> Are these changes too big to be in a minor release? Meaning: would releasing this as the next minor release break any scripts on existing DevOps piprlines that use it? Wondering if we should play it safe and see this as a breaking change. 🤔

I don't think so, all the unit and integration tests still work.

stijnmoreels (approved):
Thx for taking this up! 👍

PR #444, opened 2024-12-31
feat: implement ps script analyzer fixes

pim-simons:
Run PS Script Analyzer on build stage, implement fixes suggested by the tool.
Had to exclude [several rules](https://github.com/pim-simons/arcus.scripting/blob/feat/script-analyzer/powershell-psscriptanalyzer.psd1) to not break current functionality, we can take a look later to possible remove some exclusions.

### DIFF
--- a/powershell-psscriptanalyzer.psd1
+++ b/powershell-psscriptanalyzer.psd1
@@ -1,0 +1,15 @@
+@{
+    Severity=@('Error','Warning')
+    ExcludeRules=@('PSAvoidUsingWriteHost', 
+                   'PSUseToExportFieldsInManifest', 
+                   'PSAvoidDefaultValueForMandatoryParameter', 
+                   'PSUseApprovedVerbs',
+                   'PSUseSingularNouns',
+                   'PSUseShouldProcessForStateChangingFunctions',
+                   'PSUseDeclaredVarsMoreThanAssignments',
+                   'PSAvoidUsingPlainTextForPassword',
+                   'PSAvoidUsingUsernameAndPasswordParams',
+                   'PSReviewUnusedParameter',
+                   'PSAvoidUsingConvertToSecureStringWithPlainText'
+                   'PSAvoidGlobalVars')
+}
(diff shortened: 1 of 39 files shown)


powershell-psscriptanalyzer.psd1
```diff
@@ -0,0 +1,15 @@
+@{
+    Severity=@('Error','Warning')
+    ExcludeRules=@('PSAvoidUsingWriteHost',
```
stijnmoreels, line 3: We could look into using `Write-Verbose/Information/...` instead, as I believe that this rule is mostly there to say that not all OS systems will be able to write it 'to host'. But since most (if not all) scripts here are run in a DevOps pipeline, it might be unnecessary, indeed, to look into this.

stijnmoreels: There's also not a lot of changes in the repo itself, are we excluding too much or are we just 'that good' 😀

stijnmoreels (commented):
Thx for this DevOps improvement! ⚡

powershell-psscriptanalyzer.psd1
pim-simons: > We could look into using `Write-Verbose/Information/...` instead, as I believe that this rule is mostly there to say that not all OS systems will be able to write it 'to host'. But since most (if not all) scripts here are run in a DevOps pipeline, it might be unnecessary, indeed, to look into this.

See here for the description of the rule: https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/avoidusingwritehost?view=ps-modules#description
Since we have been using `Write-Host` since the beginning of Arcus.Scripting I don't see much reason to change this now, as just about every script is used in a DevOps pipeline indeed. 


> There's also not a lot of changes in the repo itself, are we excluding too much or are we just 'that good' 😀

What I did was fix everything that I think is relevant and are easy fixes, but exclude the checks that would break functionality. So the exclusions are a bit tailored to the code we have now. 
For example, the [PSUseApprovedVerbs](https://learn.microsoft.com/en-us/powershell/utility-modules/psscriptanalyzer/rules/useapprovedverbs?view=ps-modules) has been excluded since that would mean we would have to rename certain scripts. 
I think the goal for now is to improve the codebase where we can without any breaking changes or complete rewrites, and in the future take a good look at these exclusions and where we can change and improve further 👍

stijnmoreels (approved):

Issue #446, opened 2025-01-24
Improve documentation on DevOps related documentation


fgheysels:
When navigating to the [documentation site for Arcus.Scripting](https://scripting.arcus-azure.net/guidelines/setting-arm-outputs-to-azure-devops-variable-group/), you'll see that there's a dedicated section on how to 'set ARM outputs to DevOps variable groups':

![Image](https://github.com/user-attachments/assets/3d5e3d61-d08a-4fbd-81ae-97c714617201)

I think this is not the correct location for that part of the documentation.
Moreover, when you look into that documentation, there's some crucial information missing on how to use this particular Arcus script on Linux build-agents.   
This information however does exists in the section 'Features / Arcus DevOps'.
I think that [this page](https://scripting.arcus-azure.net/Guidelines/setting-arm-outputs-to-azure-devops-variable-group) does contain some valuable information (for instance the fact that you need to expose the 'ArmOutputs' variable explicitly as an environment variable), but that information should be merged with the documentation that exists under the 'featured' documentation [here](https://scripting.arcus-azure.net/Features/powershell/azure-devops).

PR #447, opened 2025-02-07
docs: Move and improve Devops documentation

pim-simons:
Closes https://github.com/arcus-azure/arcus.scripting/issues/446


pim-simons:
@fgheysels I moved the docs around, can you check if this is a better setup?

fgheysels (commented):
thanks for this!

pim-simons:
> thanks for this!

Any thing that needs changing or can we merge this to main?

docs/preview/03-Features/powershell/azure-devops.md
```diff
@@ -120,6 +120,92 @@ In Azure DevOps, below permissions need to be set on your variable group in orde
...
+
+      Set-AzDevOpsArmOutputsToVariableGroup -VariableGroupName 'myVariableGroup' -ArmOutputsEnvironmentVariableName 'ArmOutputs' -UpdateVariablesForCurrentJob
+```
+
+There are a few things worth noting. First of all we define `deploymentOutputs: ArmOutputs` during the `AzureResourceGroupDeployment@3` task. This means that the output parameters we specified in our Bicep template will be placed in a variable called `ArmOutputs`, this is then referenced during the execution of our script with `-ArmOutputsEnvironmentVariableName 'ArmOutputs'`.
+
+Secondly we use `-UpdateVariablesForCurrentJob` as a parameter when calling the script. This means that the output parameters from the Bicep file are also available as pipeline variables in the current running job. While not necessary in our example here, if you need to deploy another Bicep template that needs output parameters from an earlier deployed Bicep template this is the way to do it.
+
+Finally we use `system_accesstoken: $(System.AccessToken)` in the `Powershell@2` task, this is necessary because we need to use the security token used by the running build. 
+
```
fgheysels, line 200: I would add an explicit remark here that you also need to reference the ArmOutputs variable in your environment-variables section. (see line 185).

Something like this:

```suggestion
> Please note that the `ArmOutputs` variable is not available 'as is' when executing the Powershell task on a Linux agent.  When using a Linux agent, you have to explicitly add that variable in the `env:` section of the Powershell task.
```

fgheysels (approved):

fgheysels:
> > thanks for this!
> 
> Any thing that needs changing or can we merge this to main?

Looks good to me :)
I've just added a little suggestion.

PR #451, opened 2025-02-07
feat: Invoke-AzSqlDatabaseMigration supports AccessToken login

### LINKED ISSUE #450: Scripting.Sql: Support token-based access to SQL Server
The `Invoke-AzSqlDatabaseMigration` script currently only supports username/password based login to the SQL database.  
This is outdated (and should maybe even be removed in the future).
We should support token-based access as well.

fgheysels:
An additional parameter (`AccessToken`) has been added to the `Invoke-SqlDatabaseMigration` command which allows to login to a SQL Server using an access-token instead of a username/password combination.

Closes #450


pim-simons:
Is it possible to add unit/integration tests for this?

fgheysels:
> @fgheysels Is it possible to add unit/integration tests for this?

IDK.  Can we retrieve an access-token without user-interference that can be used for this ?

The tests are probably run from Azure DevOps.  Does the managed identity that is used by the DevOps pipeline has access to the SQL database that we can use in the test-suite ?

pim-simons:
> > @fgheysels Is it possible to add unit/integration tests for this?
> 
> IDK. Can we retrieve an access-token without user-interference that can be used for this ?
> 
> The tests are probably run from Azure DevOps. Does the managed identity that is used by the DevOps pipeline has access to the SQL database that we can use in the test-suite ?

Good question, probably a bit of challenge to do this without user interference. 
PR looks good to me 👍

pim-simons (approved):

PR #452, opened 2025-02-12
docs: publish v1.4.1 feature docs

pim-simons:
Publishes the v1.4.1 feature documentation.


stijnmoreels:
Normally we don't publish feature docs for fix releases, though? As it usually is not an indication that there is something different from the last minor release.

pim-simons:
> Normally we don't publish feature docs for fix releases, though? As it usually is not an indication that there is something different from the last minor release.

There are some small changes in the docs with regard to new parameters, so in this case we need to update the docs.

stijnmoreels:
> > Normally we don't publish feature docs for fix releases, though? As it usually is not an indication that there is something different from the last minor release.
> 
> There are some small changes in the docs with regard to new parameters, so in this case we need to update the docs.

Seeing the changes, I'm wondering why it is a fix release, even 🤔.

pim-simons:
> > > Normally we don't publish feature docs for fix releases, though? As it usually is not an indication that there is something different from the last minor release.
> > 
> > 
> > There are some small changes in the docs with regard to new parameters, so in this case we need to update the docs.
> 
> Seeing the changes, I'm wondering why it is a fix release, even 🤔.

I honestly just followed Frederik's suggestion of v1.4.1, since the changes are mostly his anyway.
But if it is more logical to make a 1.5.0 then I can do that as well.

stijnmoreels:
> I honestly just followed Frederik's suggestion of v1.4.1, since the changes are mostly his anyway.
But if it is more logical to make a 1.5.0 then I can do that as well.

🖐️ I would be in favor of that, yes. 😅 A minor release would be more logical in this case, I believe:
- ☑️ Contains added functionality;
- ☑️ Changes in feautre docs;
- ☑️ No issues fixed.

pim-simons:
> > I honestly just followed Frederik's suggestion of v1.4.1, since the changes are mostly his anyway.
> > But if it is more logical to make a 1.5.0 then I can do that as well.
> 
> 🖐️ I would be in favor of that, yes. 😅 A minor release would be more logical in this case, I believe:
> 
>     * ☑️ Contains added functionality;
> 
>     * ☑️ Changes in feautre docs;
> 
>     * ☑️ No issues fixed.

no worries, will do that today 👍🏻

PR #456, opened 2025-02-25
chore(docs): auto-trigger new feature docs version via release pipeline

stijnmoreels:
* Adds a GitHub workflow `bump-docs-version.yml` that runs the necessary steps described in the `/docs/README.md` to bump a new feature documentation version and proposes these changes in a PR.
* Adds a step in the Azure DevOps PowerShell Gallery release pipeline that triggers the workflow described above with the user-provided package version.

Together, they auto-trigger a new feature docs PR with the package version that is provided via Azure DevOps, making the release process more automated.

### DIFF
--- a/build/templates/trigger-bump-docs-version.yml
+++ b/build/templates/trigger-bump-docs-version.yml
@@ -16,5 +16,5 @@ steps:
       }
     displayName: 'Trigger new feature docs version PR'
     env:
-      GITHUB_TOKEN: $(gitHubToken)
+      GITHUB_TOKEN: ${{ parameters.gitHubToken }}
       VERSION: $(Build.BuildNumber)
--- a/build/templates/trigger-bump-docs-version.yml
+++ b/build/templates/trigger-bump-docs-version.yml
@@ -1,0 +1,20 @@
+parameters:
+  gitHubToken: ''
+
+steps:
+  - script: sudo apt install gh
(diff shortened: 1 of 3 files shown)


build/templates/trigger-bump-docs-version.yml
```diff
@@ -0,0 +1,20 @@
...
+
+      if ($(VERSION) -notmatch '-' -and $(VERSION) -match '^v[0-9]+\.[0-9]+\.0$') {
+        gh workflow run bump-docs-version.yml `
+          --repo arcus-azure/arcus.testing `
+          --ref main `
+          --field version=$(VERSION)
+      }
+    displayName: 'Trigger new feature docs version PR'
+    env:
+      GITHUB_TOKEN: $(gitHubToken)
```
stijnmoreels, line 19:
```suggestion
      GITHUB_TOKEN: ${{ parameters.gitHubToken }}
```

pim-simons (approved):
Looks like a great addition to me! Awesome work 👍

I don't think there is a good way to test this right, since this is a change in the powershell release pipeline?

stijnmoreels:
I checked the things separately, and also with a separate pipeline to see if the integration Azure DevOps <> GitHub works.
In any case, if it would fail when we release a new version, it won't block the release as it is the very last step, so we can than fix any issues, if there are any.

